PR #23, opened 2019-07-22
add to_list and set_type_name method

haidi-ustc:


dpdata/system.py
```diff
@@ -806,3 +858,11 @@ def check_LabeledSystem(data):
        assert( len(data['cells']) == len(data['coords']) == len(data['energies']) )
 
 
+def elements_index_map(elements,standard=False,inverse=False):
```
amcadmus: please add unit test for this function. The following cases are not covered by the current unit tests:
1, standard == False, inverse == True
2, standard == False, inverse == False 
3, standard == True, inverse = True

amcadmus: It is highly recommended to add documents that explain the parameters **standard** and **inverse**

PR #31, opened 2019-08-09
add function for read energy and force from cp2k output

robinzyb:
add function for read energy and force from cp2k output
test file in test/cp2k/ directory

from yongbin


tests/cp2k/test.py
```diff
@@ -0,0 +1,14 @@
+
+import dpdata
+
+cp2k_output = dpdata.LabeledSystem('cp2k_output', fmt = 'cp2k/output')
```
njzjz: I suggest to write a [unittest](https://docs.python.org/3/library/unittest.html)-based test framework in this file.

PR #63, opened 2019-11-29
support MultiSystems from_dir && add  more test cases for quip/gap/xyz

felix5572:
support MultiSystems from_dir && add  more test cases for quip/gap/xyz


dpdata/system.py
```diff
@@ -950,7 +950,20 @@ def from_file(cls,file_name,fmt):
         multi_systems = cls()
         multi_systems.load_systems_from_file(file_name=file_name,fmt=fmt)
         return multi_systems
-    
+
+    @classmethod
+    def from_dir(cls,dir_name, file_name, fmt='auto'):
+        multi_systems = cls()
+        target_file_list = []
+        for ii in os.walk(dir_name):
```
njzjz: Suggestion: use `glob` module to support the pattern `*`
just use
```python
target_file_list = glob.glob(os.path.join(dir_name, “**”, file_name))
```
Then `file_name` could be something like `*.xyz`, `*.log`, etc.

PR #70, opened 2019-12-21
support nopbc

njzjz:
(waiting for deepmd?)


amcadmus:
Shall we unittest for other format the default data['nopbc'] is False?

njzjz:
Done.

PR #91, opened 2020-04-07
No way to pass parameter "md" in gaussian/log

weijunji:
the gaussian/log function has a parameter `md`, but there is no way to pass it.

https://github.com/deepmodeling/dpdata/blob/2dac5ffe9432d26d4bef3d85429c1e8c884d8fae/dpdata/gaussian/log.py#L13

fix this bug, add `**kwargs` in `__init__` of `class System`

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -69,7 +69,8 @@ class System (MSONable) :
                   type_map = None,
                   begin = 0,
                   step = 1,
-                  data = None) :
+                  data = None,
+                  **kwargs) :
         """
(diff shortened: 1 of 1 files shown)


njzjz:
In fact, we can use "gaussian/md" instead of "gaussian/log", but I am not against this PR, since it makes dpdata more flexible.

————————
Jinzhe Zeng
PhD student in Chemistry
Rutgers University

jinzhe.zeng@rutgers.edu
CV: https://njzjz.win/cv

PR #109, opened 2020-08-10
Sort atom according to the type_map in Multisystem

zezhong-zhang:
1. Follow the input type_map for Multisystem instead of alphabetic order. 
2. Make the formula in the order of the type_map.

At present, the order will change when calling the check_atom_names.


amcadmus:
It is strange that python 3.6-dev passed the unit tests, but 3.5, 3.5-dev and 3.6 failed...

njzjz:
It’s the unit test that @felix5572 added.

dpdata/system.py
```diff
@@ -1460,18 +1460,17 @@ def check_atom_names(self, system):
         if len(new_in_system):
             # A new atom_name appear, add to self.atom_names
             self.atom_names.extend(new_in_system)
-            self.atom_names.sort()
```
njzjz, line 1463: I am wondering whether we need to sort atom names when self.atom_names does not have the enough element?

felix5572:
I have make a pr in your repo, this bug would be fixed. Please take a look. @zezhong-zhang

zezhong-zhang:
> I have make a pr in your repo, this bug would be fixed. Please take a look. @zezhong-zhang

Many thanks for your help, pr now merged in my repo. @felix5572

PR #115, opened 2020-09-09
dpdata label system read ase db file

jwz360:
convert ase db to dpdata label system


njzjz (changes requested):
You need to add `dpdata/ase` under this line
https://github.com/deepmodeling/dpdata/blob/f150dd1a6f01be82570bf8a5cdb4d3f1e344561a/setup.py#L27-L40

amcadmus:
documentation and unittest are required.

setup.py
```diff
@@ -33,11 +33,14 @@
...
               'dpdata/gaussian', 
               'dpdata/cp2k',
+              'dpdata/ase',
               'dpdata/xyz',
               'dpdata/pwmat', 
               'dpdata/amber',
               'dpdata/fhi_aims',
-              'dpdata/gromacs'],
+              'dpdata/gromacs',
+              'dpdata/ase'
```
njzjz, line 42: You need to use another directory name instead of `ase`, otherwise it will be conflict with the `ase` package.

njzjz (changes requested):

amcadmus:
Your change to the code should be based on branch "devel" rather than master. 

we do not want to merge some of changes on master into devel

njzjz:
Thanks for the contribution. I recommend closing this old PR and, if ASE DB-specific labeled-data support is still desired, implementing it as a small new PR against the current ASE plugin.

The current generic ASE reader can already load ASE-compatible structures, so the remaining useful scope here would mainly be extracting labels stored in `row.data`. This 2020 branch is now too far from the current codebase to merge safely: it makes ASE effectively a core import, produces an incorrect energy shape, has essentially no regression coverage, includes unrelated/generated artifacts, and conflicts with the current architecture.

A clean implementation on top of the present ASE backend would be substantially easier to review and maintain than rebasing this branch.

Agent: ChatGPT
Model: GPT-5.6 Sol

PR #116, opened 2020-10-09
Fix bug in log.py

trollchu:
For system with more than 50 atoms, gaussian will not output the " Input orientation" or "Z-Matrix orientation", so we need to add keywords "Standard orientation". Instead, you can add "geom=printinputorient" keywords in the gaussian input to get the " Input orientation".

### DIFF
--- a/dpdata/gaussian/log.py
+++ b/dpdata/gaussian/log.py
@@ -27,7 +27,7 @@ def to_system_data(file_name, md=False):
             elif line.startswith(" Center     Atomic                   Forces (Hartrees/Bohr)"):
                 flag = 1
                 forces = []
-            elif line.startswith("                          Input orientation:") or line.startswith("                         Z-Matrix orientation:"):
+            elif line.startswith("                          Input orientation:") or line.startswith("                         Z-Matrix orientation:") or line.startswith("                         Standard orientation:"):
                 flag = 5
                 coords = []
                 atom_symbols = []


njzjz (changes requested):
The standard orientation is different from the input orientation, see https://gaussian.com/symmetry/. Use the standard orientation will cause inconsistent with coordinates and forces.

trollchu:
> The standard orientation is different from the input orientation, see https://gaussian.com/symmetry/. Using the standard orientation will cause inconsistent with coordinates and forces.

Do you mean that the atom index in the force output is the same as the input orientation， but maybe not be the same as standard orientation？

njzjz:
> > The standard orientation is different from the input orientation, see https://gaussian.com/symmetry/. Using the standard orientation will cause inconsistent with coordinates and forces.
> 
> 
> 
> Do you mean that the atom index in the force output is the same as the input orientation， but maybe not be the same as standard orientation？ 

The coordinates may be rotated, but the force output will not.

trollchu:
> > > The standard orientation is different from the input orientation, see https://gaussian.com/symmetry/. Using the standard orientation will cause inconsistent with coordinates and forces.
> > 
> > 
> > Do you mean that the atom index in the force output is the same as the input orientation， but maybe not be the same as standard orientation？
> 
> The coordinates may be rotated, but the force output will not.

get it，thanks :)

Zhang-Zhiyuan-zzy:
In certain situations, it is possible that we may only require the molecular conformer without the need to align the atoms' coordinate orientation with their force vectors. In my particular case, all I need is the molecular conformers and the corresponding energies. Would it be feasible to incorporate a non-default flag to allow for this less stringent operation?

PR #121, opened 2020-12-01
ASE enhance

marian-code:
Add the ability to export energy, forces and virials to ASE atoms. It is implemented as **SinglePointCalculator** calculator which was designed to only remember results not to do any computations.

Now these methods can be used on exported **Atoms** object
```
Atoms.get_potential_energy()
Atoms.get_forces()
Atoms.get_stress()
```
The try except block in `System.to_ase_structure` method has been removed as it only catches the import error and the re-raises it and thus it provides no additional security.

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -311,11 +311,15 @@ class System (MSONable) :
             tmp.data[ii] = self.data[ii]
         
         tmp.data['cells'] = self.data['cells'][f_idx].reshape(-1, 3, 3)
-        tmp.data['virials'] = self.data['virials'][f_idx].reshape(-1, 3, 3)
         tmp.data['coords'] = self.data['coords'][f_idx].reshape(-1, self.data['coords'].shape[1], 3)
-        tmp.data['forces'] = self.data['forces'][f_idx].reshape(-1, self.data['forces'].shape[1], 3)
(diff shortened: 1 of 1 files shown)


amcadmus (changes requested):
'energies', 'forces' and 'virials' are keys supposed to appear in `LabeledSystem` rather than `System`. Please check LabeledSystem::sub_system at  https://github.com/deepmodeling/dpdata/blob/931c606d8eefbff75f7377cde828c1d084952933/dpdata/system.py#L1342

I would suggest implementing a `to_ase_structure` method in `LabeledSystem`

marian-code:
> 'energies', 'forces' and 'virials' are keys supposed to appear in `LabeledSystem` rather than `System`. Please check LabeledSystem::sub_system at
> 
> https://github.com/deepmodeling/dpdata/blob/931c606d8eefbff75f7377cde828c1d084952933/dpdata/system.py#L1342
> 
> I would suggest implementing a `to_ase_structure` method in `LabeledSystem`

I did not notice that at first, but your suggestion makes sense and is in line with class hierarchy logic. I have reimplemented `to_ase_structure` in `LabeledSystem` so it overrides the implementation in `System` that way unlabeled structures can still be exported from `System` and labeled ones are exported by overriding method in `LabeledSystem`

amcadmus (approved):

PR #127, opened 2021-02-09
fix bug in cp2k coordnates  units convert

felix5572:
In *pos-1.xyz file the coordnate units should be  Ang. But in log file, the box units are Bohr.


felix5572:
@robinzyb   Request Code  Review.

robinzyb (approved):
no problem for this pr

PR #132, opened 2021-03-16
fix bug in reading forces from OUTCAR

marian-code:
Forces are read incorrectly when `FORCE -w/o vdW` tag is output. When appropriate VASP prints this additional tag. The old read system thus read 6 force components for each atom. This failed when `LabeledSystem` tried to convert it to lower triangular
which was in this case multiplying matrices forces=(Nx6) x tranform(3x3)

```
 POSITION                                       TOTAL-FORCE (eV/Angst),                                   FORCE-w/o vdW (eV/Angst)
 --------------------------------------------------------------------------------------------------------
      1.84643     -1.87564     13.81826        -0.056448      0.108946      0.296892        -0.056448      0.053966      0.296892
      0.86824      2.31451      1.69518        -0.003983      0.006375      0.161565        -0.003983      0.006375      0.161565
      ....
```

Added test to ensure it won't happen again

### DIFF
--- a/dpdata/vasp/outcar.py
+++ b/dpdata/vasp/outcar.py
@@ -128,5 +128,5 @@ def analyze_block(lines, ntot, nelm) :
                 tmp_l = lines[jj]
                 info = [float(ss) for ss in tmp_l.split()]
                 coord.append(info[:3])
-                force.append(info[3:])
+                force.append(info[3:6])
     return coord, cell, energy, force, virial, is_converge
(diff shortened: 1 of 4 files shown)


marian-code:
The failing tests do not seem to be related to my pull request

amcadmus:
> The failing tests do not seem to be related to my pull request

Please merge with deepmodeling:master to solve the problem.  We do not test with py3.5 anymore.

PR #140, opened 2021-04-20
adding cp2k/restart_aimd_ouput

dmh1998dmh:
When a restart_aimd_output is mentioned, the first step in `log `file is skipped. A restarted cp2k aimd example was also added.


felix5572:
Could you please add a unittest test case to test the restart flag? 
@dmh1998dmh

robinzyb (approved):

PR #145, opened 2021-04-26
Add support for reading and dumping multi-frames for gro file & add dump system to gaussian input file

Ericwang6:
1. Support read from and dump to .gro file for multi-frame system
2. Add function to write gaussian input file (.gjf) with system's coordinates


dpdata/system.py
```diff
@@ -622,7 +623,7 @@ def from_deepmd_raw(self, folder, type_map = None) :
 
     @register_from_funcs.register_funcs("gro")
     @register_from_funcs.register_funcs("gromacs/gro")
-    def from_gromacs_gro(self, file_name) :
+    def from_gromacs_gro(self, file_name, format_atom_name=True) :
```
amcadmus, line 626: Could you please explain in the doc string what is the behavior of `format_atom_name=True`?

dpdata/gromacs/gro.py
```diff
@@ -1,11 +1,23 @@
 #!/usr/bin/env python3
-
+import re
 import numpy as np
+import json
+import os
 
 nm2ang = 10.
+ang2nm = 0.1
```
amcadmus, line 8: recommend to write `ang2nm=1./nm2ang`

dpdata/gaussian/gjf.py
```diff
@@ -0,0 +1,16 @@
+def to_gjf_string(system, frame_idx=0, header="", title="", foot="", charge=0, mult=1):
```
amcadmus, line 1: Please add a doc string to explain what does this method do

dpdata/system.py
```diff
@@ -631,7 +632,44 @@ def from_gromacs_gro(self, file_name) :
         file_name : str
             The input file name
         """
-        self.data = dpdata.gromacs.gro.file_to_system_data(file_name)
+        self.data = dpdata.gromacs.gro.file_to_system_data(file_name, format_atom_name=format_atom_name)
+
+    @register_to_funcs.register_funcs("gromacs/string")
```
amcadmus, line 637: The format `gromacs/string` is misleading. I would recommend to merge it into`gromacs/gro`. The interface of `gromacs/gro` can be revised to accept `file_name=None` in case one does not what to write a file, and returns the file content as a string.

dpdata/system.py
```diff
@@ -709,7 +747,56 @@ def to_pwmat_atomconfig(self, file_name, frame_idx = 0) :
         w_str = dpdata.pwmat.atomconfig.from_system_data(self.data, frame_idx)
         with open(file_name, 'w') as fp:
             fp.write(w_str)
+    
+    @register_to_funcs.register_funcs("gaussian/str")
```
amcadmus, line 751: The same issue as `gromacs/string`

amcadmus (changes requested):

tests/test_gaussian_gjf.py
```diff
@@ -0,0 +1,16 @@
...
+from context import dpdata
+
+
+class TestDumpGaussianGjf(unittest.TestCase):
+    def setUp(self):
+        self.system = dpdata.LabeledSystem('gaussian/methane.gaussianlog', 
+                                           fmt = 'gaussian/log')
+    
+    def test_dump_to_gjf(self):
+        self.system.to_gaussian_gjf("gaussian/tmp.gjf")
```
amcadmus, line 12: Do you have an expected content of the output file?

What is written in `gaussian/tmp.gjf` can be compared to the reference.

Ericwang6: Of course. But I'm not sure if it's necessary to compare them in a test case, for that the best way to check an input file is to run it with Gaussian.

amcadmus: It is necessary to ensure what the method generated is what you want.

Ericwang6: OK. I've added test case for that function.

dpdata/gaussian/gjf.py
```diff
@@ -0,0 +1,36 @@
...
```
njzjz, line 1: DP-GEN has a more complex function at https://github.com/deepmodeling/dpgen/blob/00432d2349039a3ee07817426b9aa4d7fce0f7b6/dpgen/generator/lib/gaussian.py#L98-L185. Can you take a look?

Ericwang6: According to @amcadmus 's advice, I have removed support for gaussian input files. We may later discuss how to integrate the similar function implemented in DP-GEN to dpdata. @njzjz

amcadmus (approved):

PR #149, opened 2021-05-06
Remove os.mknod in deepmd/{raw.py,comp.py}

AnguseZhang:
os.mknod method is not supported on MacOS and Windows.
See
https://stackoverflow.com/questions/32115715/os-mknod-fails-on-macos

Suggest use "with open" method

### DIFF
--- a/dpdata/deepmd/comp.py
+++ b/dpdata/deepmd/comp.py
@@ -110,5 +110,7 @@ def dump(folder,
     except OSError:
         pass
     if data.get("nopbc", False):
-        os.mknod(os.path.join(folder, "nopbc"))
+        #os.mknod(os.path.join(folder, "nopbc"))
+        with open(os.path.join(folder, "nopbc"), "w") as fw_nopbc:
+            fw_nopbc.write("")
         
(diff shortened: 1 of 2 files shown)


dpdata/deepmd/comp.py
```diff
@@ -110,5 +110,7 @@ def dump(folder,
     except OSError:
         pass
     if data.get("nopbc", False):
-        os.mknod(os.path.join(folder, "nopbc"))
+        #os.mknod(os.path.join(folder, "nopbc"))
+        with open(os.path.join(folder, "nopbc"), "w") as fw_nopbc:
+            fw_nopbc.write("")
```
njzjz, line 115: I think we don't need this line. Only a `pass` is enough.

AnguseZhang: Does it mean that "nopbc" is not supported in dump function ?

njzjz: When you call `open` with `w`, the file has already been created. So you don't need to write it.

njzjz (approved):

PR #157, opened 2021-05-18
Add support for AmberTools sqm/out and BondOrderSystem

Ericwang6:
1. Add support for AmberTools sqm output file (a quantum-chemistry software in AmberTools, mainly for AM1-BCC charge calculation)
2. Add BondOrderSystem class, support to read from .mol/.sdf file or rdkit.Chem.Mol object.


dpdata/__init__.py
```diff
@@ -9,3 +9,16 @@
...
 except ImportError:
     from .__about__ import __version__
+
+# BondOrder System has dependency on rdkit
+try:
+    import rdkit
+    USE_RDKIT = True
+except ModuleNotFoundError:
+    USE_RDKIT = False
+    print("WARNING: No rdkit is installed, 'BondOrderSystem' cannot be used")
```
njzjz, line 19: It's not a good idea to show this message to users who don't use this module.

.github/workflows/test.yml
```diff
@@ -13,12 +13,18 @@ jobs:
...
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: conda-incubator/setup-miniconda@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        auto-activate-base: true
+        activate-environment: ""
+    # install rdkit and openbabel
+    - name: Install rdkit
+      run: conda create -c conda-forge -n my-rdkit-env python=${{ matrix.python-version }} rdkit openbabel;
```
njzjz, line 24: I don't see where you use openbabel.

njzjz, line 24: You may consider to install https://pypi.org/project/rdkit-pypi/, it should be faster than conda install. Although it is not an official package, we just use it to test the package.
See rdkit/rdkit#1812.

dpdata/__init__.py
Ericwang6: OK. I'll remove this.

.github/workflows/test.yml
Ericwang6: Function `dpdata.rdkit.sanitize.convert_by_obabel()` will call obabel via command line. This is designed for the cases where the sanitization fails (using obabel to try to pre-process the mol).

Ericwang6: I'll see this. But we still have dependency on obabel, which needs conda to install.

dpdata/rdkit/sanitize.py
```diff
@@ -0,0 +1,568 @@
...
+    if not os.path.exists(cache_dir):
+        os.mkdir(cache_dir)
+    if mol.HasProp("_Name"):
+        name = mol.GetProp("_Name")
+    else:
+        name = f"mol{int(time.time())}"
+    mol_file_in = os.path.join(cache_dir, f'{name}.mol')
+    mol_file_out = os.path.join(cache_dir, f'{name}_obabel.mol')
+    Chem.MolToMolFile(mol, mol_file_in, kekulize=False)
+    out = os.popen(f"obabel {mol_file_in} -O {mol_file_out}").read()
```
njzjz, line 459: Can you consider to use openbabel's python API directly, which should have a better performance?
```suggestion
    from openbabel import openbabel
    obConversion = openbabel.OBConversion()
    obConversion.SetInAndOutFormats("mol", "mol")
    mol = openbabel.OBMol()
    obConversion.ReadFile(mol, mol_file_in)
    obConversion.WriteFile(mol, mol_file_out)
```

amcadmus (approved):

njzjz (approved):

Issue #170, opened 2021-07-09
[Feature Request] Add unit tests


njzjz:
**Summary**

Add unit tests for current modules. Some of them do not have packages installed.

**Detailed Description**

According to [Codecov](https://codecov.io/gh/deepmodeling/dpdata/tree/devel/dpdata), the following codes doesn't have unit tests.
system:
- [ ] `System.__str__`
- [ ] `System.dump`
- [ ] `System.extend`
- [x] `System.predict` (#277)
- [ ] `LabeledSystem.__str__`
- [x] `MultiSystems.__len__` (fixed by #172)
- [ ] `MultiSystems.__repr__`
- [x] `MultiSystems.__str__` (fixed by #172)
- [ ] `MultiSystems.from_dir`
- [x] `MultiSystems.get_nframes` (fixed by #172)
- [ ] `MultiSystems.predict`
- [ ] `MultiSystems.pick_atom_idx`

formats:
- [ ] `amber/md`: `from_system`
- [x] `ase/structure`: `to_system` (fixed by #171)
- [ ] `ase/structure`: `to_labeled_system`
- [x] `pymatgen/structure`: `to_system` (fixed by #171)
- [x] `pymatgen/structure`: `to_labeled_system` (fixed by #171)
- [ ] `mol`: `to_bond_order_system`
- [x] `sdf`: `to_bond_order_system` (#188)
- [ ] `siesta/output`: `from_system`
- [ ] `siesta/aimd_output`: `from_system`

PR #171, opened 2021-07-09
enable unittests of ase and pymatgen

njzjz:
related issue: #170


njzjz:
... Seems there is a breaking change in pymatgen. https://pymatgen.org/#major-announcement-v2022-0
> If your existing code uses from pymatgen import <something>, you will need to make modifications. MPRester should now be imported from pymatgen.ext.matproj. All other convenience objects such as Element, Species, Lattice, Structure, etc. should be imported from pymatgen.core.

amcadmus (approved):

PR #177, opened 2021-07-25
Support of other lammps dump coordinates 

xfanak:
The lammps dump file has 4 types of coordinate systems, two of which cannot be handled by dpdata. Support for xu, xsu were added now, with some other minor modifications.

### DIFF
--- a/dpdata/lammps/dump.py
+++ b/dpdata/lammps/dump.py
@@ -51,26 +51,23 @@ def get_natoms_vec(lines) :
     assert (sum(natoms_vec) == get_natoms(lines))
     return natoms_vec
 
-def get_coord_type(keys):
-    # Check the coordinate type in dump file
-    # 4 types in total
-    # Each coordinate type corresponds to different scale factor, to be accomplished
-    key_pc=['x','y','z'] # Plain cartesian, sf = 1
+def get_coordtype_and_scalefactor(keys):
+    # 4 types in total,with different scaling factor
+    key_pc=['x','y','z'] # plain cartesian, sf = 1
     key_uc=['xu','yu','zu'] # unwraped cartesian, sf = 1
-    key_s=['xs','ys','zs'] # scaled by lattice parameter, sf = box size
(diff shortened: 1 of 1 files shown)


dpdata/lammps/dump.py
```diff
@@ -51,29 +51,26 @@ def get_natoms_vec(lines) :
...
+    # 4 types in total,with different scaling factor
+    key_pc=['x','y','z'] # plain cartesian, sf = 1
+    key_uc=['xu','yu','zu'] # unwraped cartesian, sf = 1
+    key_s=['xs','ys','zs'] # scaled by lattice parameter, sf = lattice parameter
+    key_su = ['xsu','ysu','zsu'] #scaled and unfolded,sf = lattice parameter
+    lmp_coor_type = [key_pc,key_uc,key_su,key_sf]
+    sf = [0,0,1,1]
+    for k in range(4):
+        if all(i in keys for i in lmp_coor_type(k)):
+            return lmp_coor_type(k),sf[k]
```
njzjz, line 64: `lmp_coor_type(k)` should be `lmp_coor_type[k]`... Please test the program locally before opening the PR. Also, some new tests for new situations are required.

dpdata/lammps/dump.py
```diff
@@ -51,29 +51,26 @@ def get_natoms_vec(lines) :
...
+def get_posi(lines,coordtype) :
     blk, head = _get_block(lines, 'ATOMS')
     keys = head.split()
     id_idx = keys.index('id') - 2
-    xidx = keys.index('xs') - 2
-    yidx = keys.index('ys') - 2
-    zidx = keys.index('zs') - 2
+    xidx = keys.index(coordtype[0])-2
+    yidx = keys.index(coordtype[1])-2
+    zidx = keys.index(coordtype(2))-2
```
njzjz, line 73: Also incorrect

njzjz (changes requested):

PR #178, opened 2021-07-25
Support of all lammps dump coordinate style

xfanak:
The lammps dump file has 4 types of coordinate systems, two of which cannot be handled by dpdata. Support for xu, xsu were added now, with some other minor modifications. The codes have passed the tested on x, xs, xu, xsu styles, please let me know if other tests are needed.

### DIFF
--- a/dpdata/lammps/dump.py
+++ b/dpdata/lammps/dump.py
@@ -51,26 +51,23 @@ def get_natoms_vec(lines) :
     assert (sum(natoms_vec) == get_natoms(lines))
     return natoms_vec
 
-def get_coord_type(keys):
-    # Check the coordinate type in dump file
-    # 4 types in total
-    # Each coordinate type corresponds to different scale factor, to be accomplished
-    key_pc=['x','y','z'] # Plain cartesian, sf = 1
+def get_coordtype_and_scalefactor(keys):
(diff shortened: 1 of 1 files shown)


dpdata/lammps/dump.py
```diff
@@ -51,44 +51,38 @@ def get_natoms_vec(lines) :
...
+    key_pc=['x','y','z'] # plain cartesian, sf = 1
+    key_uc=['xu','yu','zu'] # unwraped cartesian, sf = 1
+    key_s=['xs','ys','zs'] # scaled by lattice parameter, sf = lattice parameter
+    key_su = ['xsu','ysu','zsu'] #scaled and unfolded,sf = lattice parameter
+    lmp_coor_type = [key_pc,key_uc,key_s,key_su]
+    sf = [0,0,1,1]
+    for k in range(4):
+        if all(i in keys for i in lmp_coor_type[k]):
+            return lmp_coor_type[k],sf[k]
+            break
```
njzjz, line 65: This line will never be visited, as `return` exists above.
```suggestion
```

xfanak: Got it, thank for the suggestion. Fixed.

amcadmus:
Please provide unittests for you contribution, thanks!

PR #179, opened 2021-07-26
Support of all lammps dump coordinate style

xfanak:
Corresponding tests have been added now.


amcadmus:
1. You do NOT need to close open a new PR every time you push commits. 
2. please add the UTs for the cases of "scaled" and "scaled & unfolded"

xfanak:
UTs for "scaled" and "scaled & unfolded" have been added.

dpdata/lammps/dump.py
```diff
@@ -51,44 +51,39 @@ def get_natoms_vec(lines) :
...
-    try:
-        posis = get_posi(lines) - orig
-    except ValueError:
-        fposis = get_posi_frac(lines)
-        posis = fposis @ cell
+    posis = np.array(posis)[:,1:4]
+    if sf:
+        posis = (posis%1.)@cell # convert xsu to xs first
+    else:
+        posis = (posis - orig)%np.linalg.norm(cell,axis=1)
```
amcadmus, line 86: `%np.linalg.norm(cell,axis=1)` would not work for triclinic cells.

amcadmus (changes requested):

dpdata/lammps/dump.py
xfanak: Fixed. It is %cell.sum(axis=0) now.

dpdata/lammps/dump.py
```diff
@@ -83,7 +83,7 @@ def safe_get_posi(lines,cell,orig=np.zeros(3)) :
     if sf:
         posis = (posis%1.)@cell # convert xsu to xs first
     else:
-        posis = (posis - orig)%np.linalg.norm(cell,axis=1)
+        posis = (posis - orig)%cell.sum(axis=0)
```
amcadmus, line 86: The 0th dim of cell is the frame. `cell.sum(axis=0)` sums the cell tensors of all frames. It doesn't make sense to me.

amcadmus (changes requested):

dpdata/lammps/dump.py
xfanak: The safe_get_posi function handles one frame each time. When the safe_get_posi function is called from system_data, the cell argument passed to safe_get_posi is returned from dumpbox2box function, which is always a 3*3 array.

amcadmus: I see that cell is of shape [3,3], but`%cell.sum(axis=0)` would not give the right wrapped coordinated in a triclinic cell. 

The correct way would be:
1. convert the physical coordinates to scaled coordinates. Notice that scale cell is a cubic
2. wrap the scaled coordinate by `posis%1.`
3. convert the scaled coordinates back to the physical coordinates.

xfanak: It seems that [x,y,z]@inv(cell) can accurately scale the physical coordinates, where inv is the inverse operation.

dpdata/lammps/dump.py
```diff
@@ -51,45 +51,38 @@ def get_natoms_vec(lines) :
...
-    try:
-        posis = get_posi(lines) - orig
-    except ValueError:
-        fposis = get_posi_frac(lines)
-        posis = fposis @ cell
-    return posis
+    posis = np.array(posis)[:,1:4]
+    if not sf:
+        posis = (posis-orig)@np.linalg.inv(cell)# Convert to scaled coordinates for unscaled coordinates
+    return (posis%1)@cell # Convert scaled coordinates back to Cartesien coordinates
```
amcadmus, line 85: @y1xiaoc Is it the correct behavior for scaled coordinates?

amcadmus (approved):

PR #184, opened 2021-08-03
if cp2k output is not converged, return the sys len = 0

robinzyb:


dpdata/cp2k/output.py
```diff
@@ -217,14 +217,20 @@ def get_frames (fname) :
...
     stress = []
     cell_count = 0
     coord_count = 0
+    with open(fname) as fp:
+        content = fp.read()
+        count = content.count('SCF run converged')
+        if count == 0:
+            return [], [], [], [], [], [], [], []
+
+    fp = open(fname)
```
njzjz, line 233: Instead of opening fname twice, I prefer using `fp.seek(0)`.

tests/cp2k/cp2k_output
```diff
@@ -1197,7 +1197,7 @@
 
   Total energy:                                             -1761.41773472839827
 
- *** WARNING in qs_scf.F:542 :: SCF run NOT converged ***
+ *** SCF run converged in    47 steps ***
```
robinzyb, line 1200: this file is only used to check the content from parser. add `run converged` to parse by dpdata.

njzjz (changes requested):
Can you add a test for non-coverage system like this one
https://github.com/deepmodeling/dpdata/blob/73ad1640e7511f2c2d2df8f842e529f86945168e/tests/test_gaussian_log.py#L51-L75

njzjz (approved):

PR #188, opened 2021-08-25
Bug fix of bond_order_system dump to sdf file

Ericwang6:
Bug fix of dumping to sdf file for bond_order_system and add relevant unit test cases


tests/test_bond_order_system.py
```diff
@@ -78,6 +79,16 @@ def test_dump_to_deepmd_npy(self):
                 self.assertEqual(syst['bonds'][bond_idx][ii], bonds[bond_idx][ii])
         shutil.rmtree("bond_order/methane")
     
+    def test_dump_to_sdf_file(self):
```
njzjz, line 82: A similar method `test_dump_to_mol_file` can be also added here

PR #199, opened 2021-09-12
A customized system "class CustomSystem(System)"

tuoping:
A class CustomSystem(System) is defined. Originally, it was designed to make it easier to add molecules in a periodic/in-organic lattice.
Features:
- A list `self.mols` is defined in class. It is a list of dictionaries `{"mol": mol, "at": [x, y, z], "at_is_cartisian": False}`. Here, `[x, y, z]` is the coordinates to put a molecule. `mol` is a `pymatgen.Molecule` object, which can be defined from a `.xyz` file.
- Starting with a lattice system `cs = CustomSystem("lattice.vasp", fmt="vasp/poscar")`
1. One can add a `Molecule` to the lattice by `cs.addmol(mol = mol, at = [0.0, 0.0, 0.0])`, where `mol = Molecule: obj`. Or one can add a molecule from file by `cs.addmol(file_name="molecule.xyz", at=[0.0, 0.0, 0.0])`
2. `__getitem__()` gives the data of the hybrid system, including the coords/atom_names/... of both the lattice and the molecules in one list.
3. `cs.to("vasp/poscar", "hybrid.vasp", frame_idx = m)` dumps the data of the hybrid system.
4. `replicate((n_x, n_y, n_z,))` only works with `at_is_cartisian = False`. (*To be noted: If there are more than one frames, they are supposed to share the molecule list `self.mols`. Thus, it is recommended to use relative coordinates, since cells of different frames are usually different. And molecules at some cartesian coordinates are more likely end up at unphysical positions.*)
5. The output of `__str__()` is composed of the old strings of the lattice, and a list of molecules.
6. `get_nmols()` returns the number of molecules in the hybrid system.
7. `getLatticeData()` returns the dictionary `data` with only informations of the in-organic lattice system.
8. `x = popmol(idx = -1)` removes a molecule as well as returns it.
9. **A new method `rotate_mol(self, axis = None, angle = None)`** is defined, where len(axis) == len(angle) == len(mols). Here, axis.shape == [numb_mols, 3]. When axis or angle is None, random rotation is performed on each molecule.
10. **A new method `rotate_mol_by_idx(idx, axis = None, angle = None)`** is defined, where idx points to a certain molecule in the list self.mols. Here axis.shape == 3, and angle is a float number. When axis or angle is None, random rotation is performed on molecule `self.mols[idx]`.

Added files:
class script: dpdata/customsystem.py
unittest script: tests/test_customsystem.py
unittest files: tests/hybridcrystal/*


dpdata/customsystem.py
```diff
@@ -0,0 +1,436 @@
...
+        for ii in ['atom_numbs', 'atom_names'] :
+            assert(system.data[ii] == self.data[ii])
+        for ii in ['atom_types','orig'] :
+            eq = [v1==v2 for v1,v2 in zip(system.data[ii], self.data[ii])]
+            assert(all(eq))
+        for ii in ['coords', 'cells'] :
+            self.data[ii] = np.concatenate((self.data[ii], system.data[ii]), axis = 0)
+        if self.nopbc and not system.nopbc:
+            # appended system uses PBC, cancel nopbc
+            self.data['nopbc'] = False
```
njzjz, line 192: Wring duplicated codes should be avoid. One can use
```suggestion
        super().append(system)
```
to call the parent's method.

Same for other methods.

njzjz (changes requested):

dpdata/customsystem.py
```diff
@@ -0,0 +1,436 @@
...
+def c2d(cart, cell):
+    reccell = np.zeros((3,3))
+    Vcell = np.dot(cell[0], np.cross(cell[1], cell[2]))
+    reccell[0] = np.cross(cell[1], cell[2])/Vcell
+    reccell[1] = np.cross(cell[2], cell[0])/Vcell
+    reccell[2] = np.cross(cell[0], cell[1])/Vcell
+    direct = np.zeros(3)
+    direct[0] = cart[0]*reccell[0][0] + cart[1]*reccell[1][0] + cart[2]*reccell[2][0]
+    direct[1] = cart[1]*reccell[0][1] + cart[1]*reccell[1][1] + cart[2]*reccell[2][1]
+    direct[2] = cart[2]*reccell[0][2] + cart[1]*reccell[1][2] + cart[2]*reccell[2][2]
```
njzjz, line 434: I think it's equal to
```py
    direct = (reccell.T @ cart).T
```
or something similar. Anyway, you don't need to write for every element...

njzjz: Same comment for `_merge_data`

amcadmus (commented):
I do not see any reason why we need `CustomSystem`.  It seems that the purpose of this class is to convert a `pymatgen.core.structure.Molecule` object to `dpdata.System`. They why not implementing the support of `Molecule` format for `dpdata.System`? That is one convert `Molecule` to `dpdata.System`, then do all operations on molecule with existing methods in `dpdata.System`. Is there any feature needed for `Molecule` but not supported by `dpdata.System`? 

I also do not think `CustomSystem` is a good name, because it is not self-explanatory.

tuoping:
> I do not see any reason why we need `CustomSystem`. It seems that the purpose of this class is to convert a `pymatgen.core.structure.Molecule` object to `dpdata.System`. They why not implementing the support of `Molecule` format for `dpdata.System`? That is one convert `Molecule` to `dpdata.System`, then do all operations on molecule with existing methods in `dpdata.System`. Is there any feature needed for `Molecule` but not supported by `dpdata.System`?
> 
> I also do not think `CustomSystem` is a good name, because it is not self-explanatory.

The one feature I see missing is `rotate molecule`. The main purpose of this is to 1. perturb molecules in a lattice by random rotations; 2. add molecules randomly in some of the cells in a superlattice. This can be usefull for `dpgen init_bulk`, as a complement to `perturb`. And this is usefull for organic-inorganic hybrid crystals.
- To support `Molecule` for `System`, and implement a function `rotate` in some way is certainly a better idea. (I'll try this.)
- I'm hesitating about this name as well.

dpdata/customsystem.py
tuoping: Thank you !!!

amcadmus:
One can handle one molecule by one `System`, and rotate it via `dpdata.System.affine_map`, then append the system to the rest of atoms to form the `System` you need

tuoping:
Awesome. I'll try it. Thank you.

PR #202, opened 2021-09-24
Add unit.py to manage unit conversions

Ericwang6:
1. Add unit.py to manage unit conversions
2. Fix bug of long forces in sqm.out
3. Fix bug of units in parsing forces in sqm.out


dpdata/amber/sqm.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 from ..periodic_table import ELEMENTS
 
+kcal2ev = 1 / 23.06054764
```
amcadmus, line 4: in `amber/md` a similar conversion constant is defined. How about write a common module for defining all conversion constants?

Ericwang6: Good idea! I'll develop it soon.

Ericwang6:
@amcadmus @njzjz I've add a module called `unit.py` to manage all unit conversions in `dpdata`

dpdata/unit.py
```diff
@@ -0,0 +1,104 @@
+from abc import ABC
+
+AVOGADRO = 6.02214179E23    # Avagadro constant
+ELE_CHG  = 1.602176487E-19  # Elementary Charge, in C
```
amcadmus, line 4: We may use `scipy.constants.Avogadro` and `scipy.constants.elementary_charge`.

dpdata/unit.py
```diff
@@ -0,0 +1,104 @@
...
+        self._value = lconvs[unitA] / lconvs[unitB]
+
+class ForceConversion(Conversion):
+    def __init__(self, unitA, unitB):
+        super().__init__(unitA, unitB)
+        econv = EnergyConversion(unitA.split("/")[0], unitB.split("/")[0]).value()
+        lconv = LengthConversion(unitA.split("/")[1], unitB.split("/")[1]).value()
+        self._value = econv / lconv
+
+class VirialConversion(Conversion):
```
amcadmus, line 75: I would recommend calling it `PressureConversion`.  The virial has energy unit the conversion is in another way.

amcadmus (changes requested):
great implementation, minor change required, thanks!

dpdata/unit.py
```diff
@@ -0,0 +1,104 @@
...
+            lunit = unit.split("/")[1]
+            if eunit not in econvs.keys():
+                raise RuntimeError(f"Invaild unit: {unit}")
+            if lunit not in lconvs.keys():
+                raise RuntimeError(f"Invalid unit: {unit}")
+        except:
+            raise RuntimeError(f"Invalid unit: {unit}")
+
+class Conversion(ABC):
+    def __init__(self, unitA, unitB, check=True):
```
njzjz, line 38: Can you clarify what are unitA and unitB and also add an example to this class. Others need it.

dpdata/unit.py
njzjz: What does this constant come from? As is known, some physical constants were measurements before 2019 so they are updated every four years. For example, `ELE_CHG` in 2014 is 1.602 176 6208(98) × 10−19. See [here](https://physics.nist.gov/cuu/pdf/nonsi_2014.pdf). But after the definition of some units were [changed in 2019](https://doi.org/10.1038/d41586-019-01614-8), now `ELE_CHG` is [an exact value](https://physics.nist.gov/cgi-bin/cuu/Value?evj) (1.602 176 634  x 10-19 J). I think what you use here is even older than 2014?

njzjz (changes requested):

dpdata/unit.py
njzjz: `scipy.constants.elementary_charge` uses the latest defination.

dpdata/unit.py
```diff
@@ -0,0 +1,104 @@
...
+    "kJ": 1000 / ELE_CHG
+}
+
+# length conversions
+lconvs = {
+    "angstrom": 1.0,
+    "bohr": 0.52917720859,
+    "nm": 10.0,
+    "m": 1E10,
+}
```
njzjz, line 23: We may need to confirm if `4.35974394E-18`, `2.17987197E-18`, and `0.52917720859` is the latest value or not.

amcadmus: I agree. These constants should be provided with as many digits as possible.

Ericwang6: I think we need to have a unified reference. The constant given here is from _**CRC Handbook of Chemistry and Physics**_ (90th edition), which is published in 2010. I'll look up for a latest data version

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
+from abc import ABC
+from scipy import constants
+
+AVOGADRO = constants.Avogadro                             # Avagadro constant
+ELE_CHG  = constants.elementary_charge                    # Elementary Charge, in C
+BOHR     = constants.value("atomic unit of length")       # Bohr, in m
+HARTREE  = constants.value("atomic unit of energy")       # Hartree, in Jole
+RYDBERG  = constants.Rydberg * constants.h * constants.c  # Rydberg, in Jole
```
Ericwang6, line 8: @amcadmus @njzjz  I've used constants from `scipy` package, which is the lastest

amcadmus (approved):

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+            unitA : str, unit to be converted
+            unitB : str, unit which unitA is converted to, i.e. `1 unitA = self._value unitB`
+            check : bool, whether to check unit validity
+        
+        Examples
+        --------
+        >>> conv = Conversion("foo", "bar", check=False)
+        >>> conv.set_value("10.0")
+        >>> print(conv)
+        >>> "1 foo = 10.0 bar"
```
njzjz, line 57: output value should be in this format
```suggestion
        1 foo = 10.0 bar
```

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+        >>> conv.value()
+        >>> 10.0
```
njzjz, line 59:
```suggestion
        10.0
```

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+class EnergyConversion(Conversion):
+    def __init__(self, unitA, unitB):
+        """
+        Class for energy conversion
+
+        Examples
+        --------
+        >>> conv = LengthConversion("eV", "kcal_mol")
+        >>> conv.value()
+        >>> 23.06054783061903
```
njzjz, line 89:
```suggestion
        23.06054783061903
```

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+class LengthConversion(Conversion):
+    def __init__(self, unitA, unitB):
+        """
+        Class for length conversion
+
+        Examples
+        --------
+        >>> conv = LengthConversion("angstrom", "nm")
+        >>> conv.value()
+        >>> 0.1
```
njzjz, line 103:
```suggestion
        0.1
```

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Parameters
+        ----------
+            unitA, unitB : str, in format of "energy_unit/length_unit"
+
+        Examples
+        --------
+        >>> conv = ForceConversion("kJ_mol/nm", "eV/angstrom")
+        >>> conv.value()
+        >>> 0.0010364269656262175
```
njzjz, line 121:
```suggestion
        0.0010364269656262175
```

dpdata/unit.py
```diff
@@ -0,0 +1,169 @@
...
+
+        Parameters
+        ----------
+            unitA, unitB : str, in format of "energy_unit/length_unit^3", or in `["Pa", "pa", "kPa", "kpa", "bar", "kbar"]`
+
+        Examples
+        --------
+        >>> conv = PressureConversion("kbar", "eV/angstrom^3")
+        >>> conv.value()
+        >>> 0.0006241509074460763
```
njzjz, line 141:
```suggestion
        0.0006241509074460763
```

njzjz (approved):
Approve after fixing docstring

PR #218, opened 2021-11-16
fix loading nopbc system without box

### LINKED ISSUE #215: box.raw file not found with nopbc file
I install dpdata 0.2.1 via conda.

When I tryed read the ```nobc``` example data via dpdata an error is rise.
Here belew the command I gave:
```x = dpdata.LabeledSystem("mypath/deepmd-kit/examples/nopbc/data/C1H4O2", fmt='deepmd/npy')```
Here the error:
```FileNotFoundError: [Errno 2] No such file or directory: '/mypath/deepmd-kit/examples/nopbc/data/C1H4O2/set.000/box.npy'```
Here what is written In the doc. "If one needs to train a non-periodic system, an empty ```nopbc``` file should be put under the system directory. ```box.raw``` is not necessary is a non-periodic system." 
Infact in the example directory no box.raw or virial is present.

njzjz:
Fix #215.

### DIFF
--- a/dpdata/deepmd/comp.py
+++ b/dpdata/deepmd/comp.py
@@ -8,9 +8,12 @@ def _cond_load_data(fname) :
         tmp = np.load(fname)
     return tmp
 
-def _load_set(folder) :
-    cells = np.load(os.path.join(folder, 'box.npy'))
+def _load_set(folder, nopbc: bool) :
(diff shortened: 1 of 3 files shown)


amcadmus (approved):

amcadmus:
please fix the issues in the UTs.

PR #225, opened 2021-11-19
Fix bug cp2k

trollchu:
There is a bug for cp2k output, the dpdata package cannot recognize the error when the cp2k output prints the forces multiple times. So I add an Assert to check if the shape of coords match forces. A simple test is attached.
image

### DIFF
--- a/dpdata/gaussian/log.py
+++ b/dpdata/gaussian/log.py
@@ -27,7 +27,7 @@ def to_system_data(file_name, md=False):
             elif line.startswith(" Center     Atomic                   Forces (Hartrees/Bohr)"):
                 flag = 1
                 forces = []
-            elif line.startswith("                          Input orientation:") or line.startswith("                         Z-Matrix orientation:"):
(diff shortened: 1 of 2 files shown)


dpdata/gaussian/log.py
```diff
@@ -27,7 +27,7 @@ def to_system_data(file_name, md=False):
             elif line.startswith(" Center     Atomic                   Forces (Hartrees/Bohr)"):
                 flag = 1
                 forces = []
-            elif line.startswith("                          Input orientation:") or line.startswith("                         Z-Matrix orientation:"):
+            elif line.startswith("                          Input orientation:") or line.startswith("                         Z-Matrix orientation:") or line.startswith("                         Standard orientation:"):
```
njzjz, line 30: Please revert 5fe8e19dd994bcc84ffa0ca0199bbce5421ce798 first

njzjz (changes requested):

PR #227, opened 2021-11-22
add importlib_metadata as dependency

njzjz:


AnguseZhang (changes requested):
I think such requirement is too strict and may cause conflicts, since some of our python environments in offline package  are Python3.8.

njzjz:
> I think such requirement is too strict and may cause conflicts, since some of our python environments in offline package are Python3.8.

It means it takes effect only when Python version is less than 3.8. See https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

wanghan-iapcm (approved):

PR #229, opened 2021-11-25
Fix issues in the UTs and workflow

wanghan-iapcm:
- skip the UTs on machines that do not have `rdkit`
- only mirror to gitee if the org is deepmodeling.


njzjz (approved):

tests/test_bond_order_system.py
```diff
@@ -2,12 +2,20 @@
...
 import glob
-from rdkit import Chem
-from rdkit.Chem import AllChem
+try:
+    from rdkit import Chem
+    from rdkit.Chem import AllChem
+    from dpdata import BondOrderSystem
+    skip_bond_order_system = False
+except ImportError:
+    skip_bond_order_system = True
```
njzjz, line 11: `skip_bond_order_system = False` can be put outside the `try...except` block.
```suggestion
try:
    from rdkit import Chem
    from rdkit.Chem import AllChem
    from dpdata import BondOrderSystem
except ImportError:
    skip_bond_order_system = True
else:
    skip_bond_order_system = False
```

tests/test_amber_sqm.py
```diff
@@ -4,6 +4,11 @@
 from context import dpdata
 from comp_sys import CompSys, CompLabeledSys, IsNoPBC
 
+try:
+    from dpdata import BondOrderSystem
+    skip_bond_order_system = False
+except ImportError:
+    skip_bond_order_system = True
```
njzjz, line 11:
```suggestion
try:
    from dpdata import BondOrderSystem
except ImportError:
    skip_bond_order_system = True
else:
    skip_bond_order_system = False
```

PR #269, opened 2022-04-13
generate formats table in doc

njzjz:
Also fixes some typos


njzjz:
See https://deepmodeling--269.org.readthedocs.build/projects/dpdata/en/269/formats.html

wanghan-iapcm (approved):

PR #276, opened 2022-05-08
supported non-orthogonal cells for amber/md

njzjz:
calls an external method `ase.geometry.cellpar_to_cell`. I am not sure if someone could rewrite it...


wanghan-iapcm (changes requested):
Could you please add a UT for the non-orthogonal cells?

PR #278, opened 2022-05-11
add ML labeling option for vasp ml-aimd OUTCAR

PKUfjh:
This PR intends to add labeling option for VASP machine learning force field AIMD, the default setting is the same with the original repo, which selects the first principle calculation among the MD frames as energy, force and viral labels. But one can also add option in the LabeledSystem like
```
LabeledSystem(filename,fmt,ml=True)
```
to use the machine learned energy, force and viral labels.
This option obviously is not as accurate as the default fp labels, but the motivation for adding this option (for me) is mainly to sample the structures rather than use the labeled system as training set. And you only need to change a little bit to realize the function. Also, since the default setting is the same as the original one, there is little worry to use this option in a wrong way.


njzjz:
It looks like your code is stuck in a loop.

PR #283, opened 2022-05-12
fix sqm energy

njzjz:
The code read the electronic energy (the following line of the total SCF energy) instead of the total SCF energy. This commit fixes it.

I have discussed with @Ericwang6. This should be the reason that our DP forces are good but DP energies are extremely bad.


wanghan-iapcm (approved):
Interesting bug. Approve.

Ericwang6 (approved):

PR #294, opened 2022-05-24
discriminate frames with virials labels in Multisystem class

PKUfjh:
Sometimes we came across with multisystem (such as quip/gap/xyz format files) which has both frames with virials labels and frames without virials labels. The current implementation of the Multisystems class will lead to error message:
```
RuntimeError: system has virial, but this does not
```
But this kind of Multisystems are in fact trainable, so I identify frames with virials in Multisystems, such that Systems will have names as:
```
C60  C60_virials
```
The former represents systems with 60 carbon atoms without virials labels, the latter represents systems with 60 carbon atoms with virials labels.


wanghan-iapcm (commented):
This is a breaking change to the code that has potential influence on other codes using `dpdata.MultiSystems` with explicit formula.  

I do not support the merging of this PR.

PKUfjh:
> This is a breaking change to the code that has potential influence on other codes using `dpdata.MultiSystems` with explicit formula.
> 
> I do not support the merging of this PR.

Actually this change just modifies the system names in the MultiSystems frames (add "_virials" to the system names), I do not see how this change will have negetive influence on other codes.

PR #299, opened 2022-06-04
add data type; refactor methods

njzjz:
* add DataType to manage types of data
* refactor data checking to check type and shape of data
** fix incorrect data type in some codes
* refactor methods including sub_system, append, sort_atom_types, shuffle, pick_atom_idx, and merge methods in LabeledSystem into those in System


wanghan-iapcm (approved):
brilliant idea

PR #301, opened 2022-06-08
adapt abacus/md interface to MD output without stress calculation.

Liu-RX:


wanghan-iapcm (changes requested):
Could you please provide UTs for both calc_stress == True/False cases? Thanks

Liu-RX:
> Could you please provide UTs for both calc_stress == True/False cases? Thanks

The existing UT corresponds to cal_stress == True. A new UT for cal_stress == False has already been added in the new commit.

dpdata/abacus/md.py
```diff
@@ -33,8 +33,17 @@ def get_coord_dump_freq(inlines):
...
+    if "VIRIAL" in dumplines[6]:
+        calc_stress = True
+    else:
+        assert("POSITIONS" in dumplines[6] and "FORCE" in dumplines[6])
+    nframes_dump = -1
+    if calc_stress:
+        nframes_dump = int(nlines/(total_natoms + 13))
+    else:
+        nframes_dump = int(nlines/(total_natoms + 9))
+    assert(nframes_dump > 0)
```
AnguseZhang, line 46: Could you please add explanations for the assertion when it fails? `assert nframes_dump > 0 , "The reason is xxx"`

tests/test_abacus_md.py
```diff
@@ -40,6 +50,18 @@ def test_coord(self) :
                     self.assertAlmostEqual(self.system_water.data['coords'][ii][jj][kk], coord[ii][jj][kk])
         fp.close()
 
+        fp2 = open('abacus.md.nostress/Si_coord')
```
AnguseZhang, line 53: `with open("abacus.md.nostress/Si_coord") as fp2:` is better. When the assertion `self.assertAlmostEqual` fails,  in your implementation `fp2.close()` will not be exceuted.

dpdata/abacus/md.py
AnguseZhang: Also, in `assert("POSITIONS" in dumplines[6] and "FORCE" in dumplines[6])`

AnguseZhang (approved):
Some suggestions which do not relate to the correctness.

dpdata/abacus/md.py
```diff
@@ -33,8 +33,17 @@ def get_coord_dump_freq(inlines):
...
     nlines = len(dumplines)
     total_natoms = sum(natoms)
-    nframes_dump = int(nlines/(total_natoms + 13))
-    
+    calc_stress = False
+    if "VIRIAL" in dumplines[6]:
+        calc_stress = True
+    else:
+        assert("POSITIONS" in dumplines[6] and "FORCE" in dumplines[6])
+    nframes_dump = -1
```
njzjz, line 41: This line is useless.

dpdata/abacus/md.py
Liu-RX: Thanks for the valuable suggestion! I have added assertion failed reason for each case in the abacus interface.

tests/test_abacus_md.py
Liu-RX: I have understood the problem. Now the `fp = open('...')` codes in the `test_abacus_md.py` have all been replaced with `with open('...') as fp:`. I will fix the cases in `test_abacus_scf.py` in the next pr.

dpdata/abacus/md.py
Liu-RX: Hi Jinzhe. I think the line is helpful in recognizing the keywords are in the right place of the file, so I kept the lines.

AnguseZhang: If you delete `nframes_dump`, it will still be defined in the successive conditional lines.  But I think keeping this line is also okay.

PR #304, opened 2022-06-13
add Minimizer plugin and System.minimize method

njzjz:
This PR adds the `minimize` method to `System` and also the Minimizer plugins. Also add two minimizers: ASE and AMBER SQM.


dpdata/driver.py
```diff
@@ -81,6 +83,35 @@ def label(self, data: dict) -> dict:
...
+        system = dpdata.System(data=data)
+        # list[Atoms]
+        structures = system.to_ase_structure()
+        labeled_system = dpdata.LabeledSystem()
+        for atoms in structures:
+            atoms.calc = self.ase_calculator
+            dyn = LBFGS(atoms, logfile=None)
+            dyn.run(fmax=5e-3)
+            ls = dpdata.LabeledSystem(atoms, fmt="ase/structure", type_map=data['atom_names'])
+            labeled_system.append(ls)
```
wanghan-iapcm, line 112: Does one have any chance of specifying the optimizer and the parameters for optimization?

njzjz: Good point, let me consider how to design the API

njzjz: The API is refactored.

wanghan-iapcm (commented):
Could you please provide doc on how to use the minimizer?  
especially when one wants to change the default `fmax` of "ase".

wanghan-iapcm (approved):

PR #313, opened 2022-06-18
drop conda in the test workflow

njzjz:
Here we use `pip install` instead, as it's faster than `conda install`.

On the PyPI, `rdkit` package has been released officially. `openbabel-wheel` was made by me unofficially.


njzjz:
2 minutes have been saved for each job.

wanghan-iapcm (approved):

PR #325, opened 2022-07-05
improve accuracy of sqm input coordinates

njzjz:
Coordinates with 4 digits are enough for energy calculation but not enough for force calculation or energy minimization.


wanghan-iapcm:
The UTs should be fixed before merging.

wanghan-iapcm (approved):

PR #335, opened 2022-08-02
test: update input parameters in abacus.scf UT

hongriTianqi:
update input parameters in abacus.scf UT


wanghan-iapcm (commented):
Could please explain the change of the keywords? It that due the update of abacus version? Which version are the new keywords compatible to?

hongriTianqi:
> Could you please explain the change of the keywords? It that due the update of abacus version? Which version are the new keywords compatible to?

OK. Yes, this is due to the update of abacus version. The new keywords are compatible to abacus versions up to 2.2. The changed keywords include:
atom_file => stru_file
niter        => scf_nmax
smearing => smearing_method
sigma      => smearing_sigma
force        => cal_force
stress       => cal_stress

wanghan-iapcm (approved):

PR #345, opened 2022-08-13
docs: add authors section for credits

njzjz:


njzjz:
https://deepmodeling--345.org.readthedocs.build/projects/dpdata/en/345/credits.html

PR #355, opened 2022-09-02
add option to ignore convergence check for `vasp/contcar`

panxiang126:
it is the realization for discussion #350.

### DIFF
--- a/dpdata/plugins/fhi_aims.py
+++ b/dpdata/plugins/fhi_aims.py
@@ -5,6 +5,8 @@ from dpdata.format import Format
 @Format.register("fhi_aims/output")
 class FhiMDFormat(Format):
     def from_labeled_system(self, file_name, md=True, begin = 0, step = 1, **kwargs):
+        if "converge" not in kwargs.keys():
+            kwargs["converge"] = True
         data = {}
         data['atom_names'], \
             data['atom_numbs'], \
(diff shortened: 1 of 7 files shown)


panxiang126:
Similar changes have been made to `pwmat/movement` and `fhi_aims/output` , but I don't have test files, please double-check.

wanghan-iapcm:
Please solve the issues in the UTs.

dpdata/plugins/fhi_aims.py
```diff
@@ -5,6 +5,8 @@
 @Format.register("fhi_aims/output")
 class FhiMDFormat(Format):
     def from_labeled_system(self, file_name, md=True, begin = 0, step = 1, **kwargs):
+        if "converge" not in kwargs.keys():
+            kwargs["converge"] = True
```
njzjz, line 9: This should use [`dict.setdefault`](https://docs.python.org/3/library/stdtypes.html#dict.setdefault) or be directly written into the method definition line.

panxiang126: updated.

njzjz (approved):

dpdata/plugins/fhi_aims.py
```diff
@@ -4,7 +4,7 @@
 @Format.register("fhi_aims/md")
 @Format.register("fhi_aims/output")
 class FhiMDFormat(Format):
-    def from_labeled_system(self, file_name, md=True, begin = 0, step = 1, **kwargs):
+    def from_labeled_system(self, file_name, md=True, begin = 0, step = 1, converge=True, **kwargs):
```
wanghan-iapcm, line 7: I would suggest using `req_converge` or `conv_check` in the interface, because it is more self-explanary. 
The same suggest for the pwmat and vasp interfaces.

panxiang126: All are denoted by `convergence_check`.

wanghan-iapcm (approved):

PR #368, opened 2022-10-26
fix(abacus): do not collect the unconverged structure for AIMD

pxlxingliang:


dpdata/abacus/md.py
```diff
@@ -80,6 +81,10 @@ def get_energy(outlines, ndump, dump_freq):
             if nenergy%dump_freq == 0:
                 energy.append(float(line.split()[-2]))
             nenergy+=1
+        elif "!! convergence has not been achieved" in line:
+            if nenergy%dump_freq == 0:
+                energy.append(0)
```
njzjz, line 86: If 0 is used to indicate systems that do not coverage, how can we collect 0-energy systems?

pxlxingliang: For a normal DFT calculation in ABACUS, the energy should be smaller than 0. But 0 is really not a good choice. May "None" is better. I will change it.

njzjz: Can ABACUS calculate a hydrogen ion?

njzjz: Note that `None` will make the NumPy array become an object array:
```py
>>> import numpy as np
>>> np.array([1., None]).dtype
dtype('O')
```

Maybe you can use `np.nan` or `np.inf` instead.
```py
>>> np.array([1., np.nan, np.inf]).dtype
float64
```

```py
>>> x = np.array([1., np.nan, np.inf])
>>> np.isnan(x)
array([False,  True, False])
```

njzjz:
Can you add an extra test?

dpdata/abacus/md.py
pxlxingliang: > Can ABACUS calculate a hydrogen ion?

Yes. I see. The previous statement is indeed incorrect.

pxlxingliang: > Note that `None` will make the NumPy array become an object array:
> 
> ```python
> >>> import numpy as np
> >>> np.array([1., None]).dtype
> dtype('O')
> ```
> 
> Maybe you can use `np.nan` or `np.inf` instead.
> 
> ```python
> >>> np.array([1., np.nan, np.inf]).dtype
> float64
> ```
> 
>  ```python
> >>> x = np.array([1., np.nan, np.inf])
> >>> np.isnan(x)
> array([False,  True, False])
> ```

Thanks for your suggestion, I don't know about the object array before.
I will change to "np.nan"

pxlxingliang:
> Can you add an extra test?

Ok, I will add one.

njzjz (approved):

PR #377, opened 2022-11-22
fix bug in reading trajs with random tid

Angel-Jia:
fix #376


dpdata/lammps/dump.py
```diff
@@ -32,7 +43,8 @@ def get_atype(lines, type_idx_zero = False) :
     atype = []
     for ii in blk :
         atype.append([int(ii.split()[id_idx]), int(ii.split()[tidx])])
```
wanghan-iapcm, line 45:
```suggestion
        atype.append([int(ii.split()[tid]), int(ii.split()[id_idx])])
```

dpdata/lammps/dump.py
```diff
@@ -76,16 +88,17 @@ def safe_get_posi(lines,cell,orig=np.zeros(3), unwrap=False) :
...
     yidx = keys.index(coordtype[1])-2
     zidx = keys.index(coordtype[2])-2
     sel = (xidx, yidx, zidx)
     posis = []
     for ii in blk :
         words = ii.split()
-        posis.append([float(words[id_idx]), float(words[xidx]), float(words[yidx]), float(words[zidx])])
-    posis.sort()
-    posis = np.array(posis)[:,1:4]
+        posis.append([float(words[id_idx]), float(words[tidx]), float(words[xidx]), float(words[yidx]), float(words[zidx])])
```
wanghan-iapcm, line 99:
```suggestion
        posis.append([float(words[tidx]), float(words[id_idx]), float(words[xidx]), float(words[yidx]), float(words[zidx])])
```
Then we do not need to implement _sort_tid_comparator

wanghan-iapcm (commented):
I suggest an easier way of implementing the sorting according to type. Please check if I made anything wrong.

dpdata/lammps/dump.py
Angel-Jia: `_sort_tid_comparator` is used so that there is only one possibility for sorting.
In this case, we may get multiple orders when sorting. For example:
```
# order1
id  type_id        x      y     z
1       1            1.0   1.0   1.0
2       1            2.0   2.0   2.0
3       2            3.0   2.0   2.0

# order2
id  type_id        x      y     z
2       1            2.0   2.0   2.0
1       1            1.0   1.0   1.0
3       2            3.0   2.0   2.0
```
Would this be OK?

wanghan-iapcm: Please check:
```
In [18]: a = np.array([[1, 2], [1, 1]])                                                                                                                                               

In [19]: a.sort(axis=0)                                                                                                                                                               

In [20]: a                                                                                                                                                                            
Out[20]: 
array([[1, 1],
       [1, 2]])
```

wanghan-iapcm: please check
```
In [33]: a = [[1, 2, 1], [1, 1, 3]]                                                                                                                                                   

In [34]: a.sort()                                                                                                                                                                     

In [35]: print(a)                                                                                                                                                                     
[[1, 1, 3], [1, 2, 1]]

```

Angel-Jia:
Complete!

wanghan-iapcm (approved):

PR #383, opened 2022-11-29
fix issue #381. 

amcadmus:
The returned system should be of the same type as the perturbed system

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -810,7 +810,12 @@ class System (MSONable) :
         perturbed_system : System
             The perturbed_system. It contains `pert_num` * frame_num of the input system frames.
         """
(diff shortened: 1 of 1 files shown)


njzjz (approved):
Approve, but at this time, the energies and forces will not match the coordinates.

wanghan-iapcm:
> Approve, but at this time, the energies and forces will not match the coordinates.

You are right. what is the best behavior? raise an exception?

njzjz:
Sorry I think this PR does not work. The reason to cause this error is that the shape of energies and forces do not match the shape of coordinates. So only coordinates should be added into data, or just raise RuntimeError.

njzjz (approved):

PR #413, opened 2023-01-24
lint and automatic lint

### LINKED ISSUE #344: Code format
**Summary**

Do we have a general python format guideline (i.e. Black)?  I noticed some format difference between files. If I use a format plugin in vscode, the actually code changes can be mixed with the format changes which can be confusing. Maybe we can fix the format issue in one commit and use a format workflow for all the following codes submitted.

njzjz:
Fix #344.

To lint codes, this PR uses a popular tool, [pre-commit](https://pre-commit.com/). `.pre-commit-config.yaml` shows all hooks used, including [black](https://github.com/psf/black) and some common fixes. It can add a git hook to `git commit`. Thus, everything committed can be already formatted. In addition, there is also a [CI](https://pre-commit.ci/) to fix PRs. This way, we can ensure that everything new will also be formatted.

This PR introduces a huge change history to almost all files, but it's a necessary step to start linting our codes.


njzjz:
The CI also fixes this PR itself.

Screenshot from 2023-01-24 18-17-02

njzjz:
@wanghan-iapcm if you think this resolution is ok, I also want to apply it to other projects.

wanghan-iapcm (approved):
It is a nice solution!

PR #422, opened 2023-02-19
Add support for 'deepmd/mixed' format with dpdata.MultiSystems

iProzd:
Add support for 'deepmd/mixed' format with dpdata.MultiSystems
1. Support dump from dpdata.MultiSystems to [mixed type format](https://github.com/deepmodeling/deepmd-kit/blob/master/doc/model/train-se-atten.md#data-format):
    dpdata.MultiSystems.to_deepmd_mixed('dir_name_mixed')
2. Support load from mixed type format to dpdata.MultiSystems:
    dpdata.MultiSystems.load_systems_from_file('dir_name_mixed', fmt='deepmd/mixed')

### DIFF
--- a/dpdata/plugins/deepmd.py
+++ b/dpdata/plugins/deepmd.py
@@ -7,6 +7,8 @@ import dpdata
 import dpdata.deepmd.comp
 import dpdata.deepmd.hdf5
 import dpdata.deepmd.raw
+import dpdata.deepmd.mixed
+import os
 from dpdata.driver import Driver
 from dpdata.format import Format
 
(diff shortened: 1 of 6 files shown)


iProzd:
I've add unittest for this, but I'm wondering where to add the docs.

iProzd:
@wanghan-iapcm @njzjz

njzjz:
The docs can be automatically generated in https://deepmodeling--422.org.readthedocs.build/projects/dpdata/en/422/formats.html

dpdata/plugins/deepmd.py
```diff
@@ -69,6 +71,88 @@ def from_labeled_system(self, file_name, type_map=None, **kwargs):
     MultiMode = Format.MultiModes.Directory
 
 
+@Format.register("deepmd/mixed")
```
njzjz, line 74: I think it's actually `deepmd/npy/mixed`. We can also have `deepmd/raw/mixed` and `deepmd/hdf5/mixed`.

dpdata/plugins/deepmd.py
```diff
@@ -69,6 +71,88 @@ def from_labeled_system(self, file_name, type_map=None, **kwargs):
...
+class DeePMDMixedFormat(Format):
+    def from_system_mix(self, file_name, type_map=None, **kwargs):
+        return dpdata.deepmd.mixed.to_system_data(
+            file_name, type_map=type_map, labels=False
+        )
+
+    def to_system(self, data, file_name, set_size=200, prec=np.float64, **kwargs):
```
njzjz, line 81: Is there any reason to set the default of set size to 200? If only one set is supported, I think we should remove this option.

dpdata/plugins/deepmd.py
```diff
@@ -69,6 +71,88 @@ def from_labeled_system(self, file_name, type_map=None, **kwargs):
...
+        return dpdata.deepmd.mixed.to_system_data(
+            file_name, type_map=type_map, labels=True
+        )
+
+    def mix_system(self, *system, **kwargs):
+        """Mix the systems into mixed_type ones
+
+        Parameters
+        ----------
+        file_name : str
```
njzjz, line 111: This doc doesn't match the actual parameters.

dpdata/system.py
```diff
@@ -558,6 +562,33 @@ def append(self, system):
...
+        if type_map is None:
+            type_map = self.get_atom_names()
+        type_index = [type_map.index(i) for i in self.data["atom_names"]]
+        frames = self.get_nframes()
+        self.data["real_atom_types"] = np.tile(
+            np.array([type_index[i] for i in self.data["atom_types"]]), [frames, 1]
+        )
+        self.data["real_atom_names"] = type_map
+        natoms = self.get_natoms()
+        self.data["atom_types"] = np.array([0 for i in range(natoms)])
```
njzjz, line 588: `np.zeros((natoms,), dtype=int)`

wanghan-iapcm (changes requested):
I find the UT is trivial: All systems has the same types and numbers of atoms. Please provide tests for the cases 
- systems has the same total number of atoms but different atom types, 
- systems has different total numbers of atoms,

and  see if your code is working as it was designed. 

The doc for the mix type is also missing. Please provide the doc.

dpdata/plugins/deepmd.py
```diff
@@ -69,6 +71,92 @@ def from_labeled_system(self, file_name, type_map=None, **kwargs):
     MultiMode = Format.MultiModes.Directory
 
 
+@Format.register("deepmd/npy/mixed")
+class DeePMDMixedFormat(Format):
```
njzjz, line 75: Could you add documentation here for this format?
(I am not sure what is the best way to do it, but we should have an automatic way to have documentation for so many different formats)

njzjz (changes requested):

njzjz (approved):

wanghan-iapcm (approved):

PR #424, opened 2023-02-23
Correct the headers in README.md

iProzd:


njzjz:
The previous header is correct. However, other headers are wrong. In a doc, there should be only one `= title =` (i.e. h1).

iProzd:
> The previous header is correct. However, other headers are wrong. In a doc, there should be only one `= title =` (i.e. h1).

Then maybe I also correct other headers in this PR?

njzjz (approved):

PR #425, opened 2023-02-24
fix: standardize the deepmd/npy/mixed format

iProzd:
This PR has concated two commits together:

1. Update the dpdata.MultiSystems() when from_deepmd_npy_mixed method is called;

dpdata.MultiSystems().from_deepmd_npy_mixed only returned the results before but did not change itself, which is fixed in this commit, to be consistent with other from methods. 
(another bug is also fixed: not using .copy() in data["atom_names"] may cause error when manually changing type_map for this system. UTs are added in the next commit.)

2. Allow multiple sets in mixed-type format;

Now for maximum 50000 frames in one sys and 2000 frames in one set.
The reason I did not use 5000 frames per set, is that I think maximum set frames will be much more often used in mixed-type format than other format, and 2000 will be enough for large batch and more friendly for memory.

Add UTs for type_map changing and mixed_type dir check.

### DIFF
--- a/dpdata/deepmd/mixed.py
+++ b/dpdata/deepmd/mixed.py
@@ -54,61 +54,80 @@ def to_system_data(folder, type_map=None, labels=True):
     if os.path.isfile(os.path.join(folder, "nopbc")):
         data["nopbc"] = True
     sets = sorted(glob.glob(os.path.join(folder, "set.*")))
-    assert len(sets) == 1, "Mixed type must have only one set!"
-    cells, coords, eners, forces, virs, real_atom_types = _load_set(
-        sets[0], data.get("nopbc", False)
-    )
-    nframes = np.reshape(cells, [-1, 3, 3]).shape[0]
-    cells = np.reshape(cells, [nframes, 3, 3])
-    coords = np.reshape(coords, [nframes, -1, 3])
-    real_atom_types = np.reshape(real_atom_types, [nframes, -1])
-    natom = real_atom_types.shape[1]
-    if labels:
-        if eners is not None and eners.size > 0:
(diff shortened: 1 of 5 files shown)


wanghan-iapcm (changes requested):
Please add the UTs that can detect the two fixed bugs.

dpdata/deepmd/mixed.py
```diff
@@ -197,12 +226,12 @@ def mix_system(*system, type_map, split_num=200, **kwargs):
...
         Maps atom type to name
     split_num : int
-        Number of frames in each system
+        Number of max frames in each system
 
     Returns
     -------
     mixed_systems: dict
-        dict of mixed system with key '{atom_numbs}/sys.xxx'
+        dict of mixed system with key 'atom_numbs.sys.xxx'
```
wanghan-iapcm, line 234: why not merging all systems with the same number of atoms into one system?

dpdata/deepmd/mixed.py
```diff
@@ -227,7 +256,7 @@ def mix_system(*system, type_map, split_num=200, **kwargs):
                     temp_systems[str(natom)], split_num=split_num
                 )
                 sys_name = (
-                    f"{str(natom)}/sys." + "%.6d" % atom_numbs_sys_index[str(natom)]
+                    f"{str(natom)}.sys." + "%.3d" % atom_numbs_sys_index[str(natom)]
```
wanghan-iapcm, line 259: is %3d enough?

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #432, opened 2023-03-12
Raise error when from_deepmd_npy_mixed with not-a-directory para.

iProzd:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when importing mixed-format data from an invalid path.
  * A clear `FileNotFoundError` is now raised when the supplied path is not a directory.

### DIFF
--- a/dpdata/plugins/deepmd.py
+++ b/dpdata/plugins/deepmd.py
@@ -136,7 +136,7 @@ class DeePMDMixedFormat(Format):
 
     def from_multi_systems(self, directory, **kwargs):
         sys_dir = []
-        assert os.path.isdir(directory), f'{directory} is not a directory!'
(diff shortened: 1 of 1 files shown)


dpdata/plugins/deepmd.py
```diff
@@ -136,6 +136,7 @@ def mix_system(self, *system, type_map, **kwargs):
 
     def from_multi_systems(self, directory, **kwargs):
         sys_dir = []
+        assert os.path.isdir(directory), f"{directory} is not a directory!"
```
wanghan-iapcm, line 139: Please raise `FileNotFoundError`. One has not chance of catching the error with assertion! 
Plz also add a UT for it.

wanghan-iapcm (changes requested):

njzjz (approved):

PR #450, opened 2023-04-12
fix(abacus): fix bug in read LATTICE_CONSTANT in md

pxlxingliang:
In ABACUS version >= v3.1.4, the format of file "MD_dump" is different. The lattice constant line is added by the unit "Angstrom" at the end, and the position is the real cartesian coordinate. 
MD_dump is changed from 
```
MDSTEP:  0
LATTICE_CONSTANT: 23.453780000000
LATTICE_VECTORS
  1.000000000000  0.000000000000  0.000000000000
  0.000000000000  1.000000000000  0.000000000000
  0.000000000000  0.000000000000  1.000000000000
VIRIAL (KBAR)
  36.689617311102  0.000000000001  -0.000000000000
  0.000000000001  36.689617311104  -0.000000000001
  -0.000000000000  -0.000000000001  36.689617311147
INDEX    LABEL    POSITIONS    FORCE (eV/Angstrom)
  0  Sn  0.000000000000  0.000000000000  0.000000000000  -0.000000000000  0.000000000000  -0.000000000000
  1  Sn  0.000000000000  0.250000000000  0.250000000000  -0.000186795145  -0.000453823732  -0.000453823732
```

to

```
MDSTEP:  0
LATTICE_CONSTANT: 12.411200939060 Angstrom
LATTICE_VECTORS
  1.000000000000  0.000000000000  0.000000000000
  0.000000000000  1.000000000000  0.000000000000
  0.000000000000  0.000000000000  1.000000000000
VIRIAL (kbar)
  36.689617311101  0.000000000003  -0.000000000000
  0.000000000003  36.689617311089  -0.000000000001
  -0.000000000000  -0.000000000001  36.689617311135
INDEX    LABEL    POSITION (Angstrom)    FORCE (eV/Angstrom)    VELOCITY (Angstrom/fs)
  0  Sn  0.000000000000  0.000000000000  0.000000000000  -0.000000000000  -0.000000000001  -0.000000000001  0.001244557166  -0.000346684288  0.000768457739
  1  Sn  0.000000000000  3.102800034079  3.102800034079  -0.000186795145  -0.000453823768  -0.000453823768  0.000550996187  -0.000886442775  0.001579501983
```

PR #461, opened 2023-04-27
reading the MD_dump file of ABACUS3.2

lauthirteen:
The MD_dump file format for ABACUS3.2 has changed


pxlxingliang:
Hi, @lauthirteen , this bug has been fixed in #450

PR #466, opened 2023-05-08
fix bug in reading cp2k/output of cp2k v9.1 (compatible with old ver)

ChiahsinChu:


njzjz:
xref: https://github.com/readthedocs/readthedocs.org/issues/10290 for failed docs build

robinzyb:
This seems no unittest output present for 9.1 version.

PR #472, opened 2023-05-18
add new format : qe/pw/md

pee8379:
allow dpdata accept several new calculation modes of quantum espresso pw; md, vc-md, relax, vc-relax,

new format key 'qe/pw/md' is used.

many of codes are derived from original scf.py file.


wanghan-iapcm (changes requested):
Could you please add unittests for your PR? Thank you.

PR #476, opened 2023-05-23
allow custom data types for deepmd format

njzjz:
In this commit, the deepmd format reads data types from `dpdata.system.LabeledSystem.DTYPES` and load or dump the corresponding data. One can modify `dpdata.system.LabeledSystem.DTYPES` before using it or in a plugin (see tests), so deepmd format can be extended in this way.

### DIFF
--- a/dpdata/deepmd/comp.py
+++ b/dpdata/deepmd/comp.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 
 import dpdata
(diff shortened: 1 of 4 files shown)


njzjz:
@robinzyb this PR should be also useful to you if you want to implement any new data types.

wanghan-iapcm (approved):

PR #511, opened 2023-07-26
new feature DFTB+

ruio248:


dpdata/dftb_plus/inout.py
```diff
@@ -0,0 +1,45 @@
...
+                forces.append([float(s[1]), float(s[2]), float(s[3])])
+                flag += 1
+                if flag == 12:
+                    flag = 0
+            elif line.startswith("Total energy:"):
+                s = line.split()
+                energy = float(s[2])
+
+    symbols = np.array(symbols)
+    forces = -np.array(forces)
```
njzjz, line 41: It seems that DFTB+ outputs forces instead of gradients, so forces should not multiply by -1.

(need check)

dpdata/dftb_plus/inout.py
```diff
@@ -0,0 +1,45 @@
+from typing import Tuple
+
+import numpy as np
+
+
+def read_dftb_plus(fn: str) -> Tuple[str, np.ndarray, float, np.ndarray]:
```
njzjz, line 6: DFTB+ has input and output files separately, so I suggest not merging them into one file. Instead, pass a directory or two file arguments.

njzjz (changes requested):

wanghan-iapcm (commented):
Please provide doc for this format.

tests/dftb_plus/dftb_pin.hsd
```diff
@@ -0,0 +1,93 @@
+Geometry = GenFormat {
+4 C
+N H
+1     1        1.014150        0.112320        0.047370
+2     2        3.909390        0.037985       -0.101159
+3     2        0.702550       -0.851820       -0.060860
+4     2        0.702550        0.603740       -0.789160
+}
```
njzjz, line 8: From dftb+ documentation

> The second line contains the chemical symbols of the elements present separated by one or more spaces. The following n lines contain a list of the atoms. The first number is the atom number in the structure (this is currently ignored by the program). The second number is the chemical type from the list of symbols on line 2. Then follow the coordinates. For S and C format, these are x, y, z in Å, but for F they are fractions of the three lattice vectors.

So the unit is Angstrom

ruio248:
> Please provide doc for this format.
The doc is provided in dpdata/plugins/dftbplus.py

njzjz (approved):

wanghan-iapcm (commented):
Please update dftb+ format to readme:

Issue #517, opened 2023-08-03
[BUG] Cell file read incorrectly for 'qe/cp/traj' format


rashidrafeek:
When reading a `qe/cp/traj` output, the cells read from the `[prefix].cel` file is read incorrectly. The cells written in the `[prefix].cel` file produced in a cp.x calculation is transposed as mentioned [here](https://www.quantum-espresso.org/Doc/cp_user_guide/cp_user_guide.html#output-files). But while reading them they are not transposed by dpdata and thereby a transposed cell instead of the original cell is read by dpdata. I noticed this when I tried to use dpdata to read a trajectory I ran previously and used the `to_ase_structure` function of `LabeledSystem` to write it into a different format and visualized it. I think the [test file](https://github.com/deepmodeling/dpdata/blob/85a3b5eb40a2fae63dd87770d0927b5bc8706241/tests/qe.traj/traj6.cel) which tests this functionality was not created by a quantum espresso cp.x calculation.

I have modified dpdata locally to fix this and have confirmed that transposing the cell gives the desired result by visualizing. I can submit a pull request fixing this if it is okay.

rashidrafeek:
Yeah, I have tested the PR and it does fix this issue. And yes, the transposed test file in the PR is similar to the one produced by QE.

PR #520, opened 2023-08-07
Fix issue #517

wanghan-iapcm:


rashidrafeek:
I have tested this and it does fix the issue. Thanks for the fix.

PR #541, opened 2023-09-14
replace the old data type with the same name and throw warning

njzjz:
When a data type is unexpectedly registered twice, the behavior of some methods will be strange (for example, `append`).

For this reason, the old one is replaced, and a warning is thrown.


wanghan-iapcm (commented):
Could you please provide a UT for this situation?

PR #550, opened 2023-10-11
avoid formula causing file name too long

zjgemi:


njzjz:
This change is breaking.

wanghan-iapcm:
close this issue because #554 has fixed it.

PR #569, opened 2023-11-02
raise error instead of sys.exit

njzjz:
`sys.exit` is a lousy implementation when one wants to catch the error.


njzjz:
ping @wanghan-iapcm

wanghan-iapcm (approved):

PR #583, opened 2023-11-30
Enable loading of virials for 'qe/cp/traj' if prefix.str file is present

rashidrafeek:
Currently the virials are not read when reading a QE cp.x MD. But the stresses are printed in the `prefix.str` by default. This PR adds support to load virials if this file is present.

### DIFF
--- a/dpdata/plugins/qe.py
+++ b/dpdata/plugins/qe.py
@@ -1,8 +1,9 @@
+import numpy as np
+
 import dpdata.md.pbc
 import dpdata.qe.scf
 import dpdata.qe.traj
 from dpdata.format import Format
-import numpy as np
 
 
 @Format.register("qe/cp/traj")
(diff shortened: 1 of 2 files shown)


njzjz (changes requested):
Could you add a test for virial?

https://github.com/deepmodeling/dpdata/blob/e94866144d13730b8244a671177d940bf8e4ef2c/tests/test_qe_cp_traj.py#L9

njzjz:
Thanks for the contribution. I recommend closing this PR rather than reviving the current branch.

The original goal—loading virials for `qe/cp/traj` from the `.str` file—has since been implemented more completely in #859 / `beee99a`. The current implementation validates `.pos` / `.str` step alignment, reports mismatches with clearer diagnostics, computes virials with the intended units, and includes regression tests for both successful and failing cases.

This PR also currently conflicts with `master` and still lacks the virial regression test requested in the earlier review. Given that the functionality has already been superseded, rebasing and maintaining this older implementation would add little value.

Agent: ChatGPT
Model: GPT-5.6 Sol

PR #597, opened 2024-01-21
Orca support

hl2500:
Added support for ORCA single point energy files


njzjz:
@CodiumAI-Agent /review --pr_reviewer.inline_code_comments=true

njzjz (approved):
Looks good to me

wanghan-iapcm:
@njzjz shall we remove the supported format table from the README and relies only on the automated format doc?

njzjz:
> @njzjz shall we remove the supported format table from the README and relies only on the automated format doc? 

I agree. Indeed, we also need to merge other parts.

PR #603, opened 2024-02-01
fix invalid condition in compLabeledSys in unittest

robinzyb:
```python
"virials" not in self.system_1
```
always returns `True`. 

Quick check by inserting and run unittest:
```python

        #     self.assertEqual(len(self.system_1['virials']), 0)
        #     return
        print("valid condition:", not self.system_1.has_virial())
        print("current condition:", "virials" not in self.system_1)
        if not self.system_1.has_virial():
            self.assertFalse(self.system_2.has_virial())
            return
        np.testing.assert_almost_equal(
            self.system_1["virials"],
```

### DIFF
--- a/dpdata/driver.py
+++ b/dpdata/driver.py
@@ -163,6 +163,8 @@ class HybridDriver(Driver):
             else:
                 labeled_data["energies"] += lb_data["energies"]
                 labeled_data["forces"] += lb_data["forces"]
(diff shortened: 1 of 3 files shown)


robinzyb:
Failed tests should be carefully examed. Please contact corresponding maintainers.

njzjz (approved):
I fix the tests in #604

PR #606, opened 2024-02-06
Add unit conversion and nopbc occasion for quip/gap/xyz

SchrodingersCattt:
- Common units of energy and force are converted to eV and eV/angstrom to adapt datasets in such format. 
- The "nopbc" is applied when Lattice information does not exist.


wanghan-iapcm (changes requested):
could you please add UTs for the implementation? Thanks!

njzjz:
The old UT doesn't pass.

PR #614, opened 2024-03-13
add ASE's traj support

thangckt:
add [ASE's traj](https://wiki.fysik.dtu.dk/ase/ase/io/trajectory.html) support


dpdata/plugins/ase.py
njzjz, line 1: When I look at this file, it seems to be in a mass... `ASEStructureFormat.from_system` tries to read files from an `ase.Atoms` but `ASEStructureFormat.from_multi_systems` tries to read files from a file.

It doesn't need to be resolved in this PR.

dpdata/plugins/ase.py
```diff
@@ -187,6 +188,135 @@ def to_labeled_system(self, data, *args, **kwargs):
         return structures
 
 
+@Format.register("ase/traj")
+class ASETrajFormat(Format):
```
njzjz, line 192: A unit test should be added for this test. I found there have already been testing cases in `tests/test_ase_traj.py` and related trajectory files, so just writing in that file is okay.

dpdata/plugins/ase.py
```diff
@@ -187,6 +188,135 @@ def to_labeled_system(self, data, *args, **kwargs):
...
+            "atom_names": None,
+            "atom_numbs": None,
+            "atom_types": None,
+            "orig": None,
+            "nopbc": None,
+            "cells": [None] * len(sub_traj),
+            "coords": [None] * len(sub_traj),
+            "energies": [None] * len(sub_traj),
+            "forces": [None] * len(sub_traj),
+            "virials": [None] * len(sub_traj),
```
njzjz, line 293: This might cause errors (not sure). Need a unit test for the virial not existing case.

njzjz (changes requested):

dpdata/plugins/ase.py
thangckt: not related to this PR

dpdata/plugins/ase.py
thangckt: I am not familiar with unit test, however I tried to put a simple test in that file.
Please check it

dpdata/plugins/ase.py
thangckt: I put a new unlabeled traj file for this test

thangckt: why error?

thangckt: I think lines 305 to 310  can prevent this

tests/test_ase_traj.py
```diff
@@ -43,5 +43,16 @@ def setUp(self):
         self.v_places = 4
 
 
+@unittest.skipIf(skip_ase, "skip ase related test. install ase to fix")
+class TestASEtraj3(unittest.TestCase, CompLabeledSys, IsPBC):
+    def setUp(self):
+        self.system_1 = dpdata.System("ase_traj/Cu.traj", fmt="ase/traj")
+        self.system_2 = dpdata.LabeledSystem("ase_traj/Cu.traj", fmt="ase/traj")
```
njzjz, line 50: I've seen the error message in the log. When using `comp_sys.CompLabeledSys`, it's expected to provide two `LabeledSystem`. `comp_sys.CompSys` compares two `System`.

thangckt: so, what i have to do?

thangckt: so what should I do?

tests/test_ase_traj.py
```diff
@@ -48,6 +48,8 @@ class TestASEtraj3(unittest.TestCase, CompLabeledSys, IsPBC):
     def setUp(self):
         self.system_1 = dpdata.System("ase_traj/Cu.traj", fmt="ase/traj")
         self.system_2 = dpdata.LabeledSystem("ase_traj/Cu.traj", fmt="ase/traj")
```
robinzyb, line 50: the two systems are used to compare the outcomes of your code. 
So the first one is parsing from the `ase_traj`, 
```
self.system_1 = dpdata.LabeledSystem("ase_traj/Cu.traj", fmt="ase/traj")
```
the second one is the numerical outcomes stored as deepmd format. 
self.system_1 = dpdata.LabeledSystem("results you expect to be", fmt="deepmd/npy")

thangckt:
There is an error when test with python 3.7
```
AssertionError: Lists differ: ['Mo', 'S'] != ['S', 'Mo']
```
It may originate at these lines
image
The order of atom-names is correct with python 3.8 3.9 and on my machine with 3.11, but fail with python 3.7

It is the problem with the old code, but not due to this PR
Please check it

thangckt:
> There is an error when test with python 3.7
> 
> ```
> AssertionError: Lists differ: ['Mo', 'S'] != ['S', 'Mo']
> ```
> 
> It may originate at these lines image The order of atom-names is correct with python 3.8 3.9 and on my machine with 3.11, but fail with python 3.7
> 
> This is the problem of the existed code, but not due to this PR Please check it

thangckt:
> It may originate at these lines image The order of atom-names is correct with python 3.8 3.9 and on my machine with 3.11, but fail with python 3.7
> 
> This is the problem of the existed code, but not due to this PR Please check it

njzjz (approved):
I add a test for the unlabeled system. Then I think it's good to merge.

robinzyb (approved):

PR #619, opened 2024-03-19
Make improvements to the format for OpenMX

shigeandtomo:
Dear developers,

I would like to make some improvements on the format for OpenMX.

1.  Change the format name from `openmx/out` to `openmx/md`
2. Add the if statement to read geometry optimization trajectories
3. Add the if statement to check if the SCF calculation has converged

Thank you in advance.


dpdata/plugins/openmx.py
```diff
@@ -3,15 +3,15 @@
 from dpdata.format import Format
 
 
-@Format.register("openmx/out")
+@Format.register("openmx/md")
```
njzjz, line 6: This might be good, as the old one has never been released.

njzjz (approved):

PR #626, opened 2024-04-02
PyMatgenStructureFormat with from_system method

MoseyQAQ:
Now the PyMatgenStructureFormat doesn't support the <kbd>from_system</kbd> method. 
This PR adds this method, enabling users to convert the <kbd>pymatgen.core.Structure</kbd> to the dpdata format, which is convenient when you use <kbd>pymatgen.transformations</kbd> to build the atomic model.


njzjz (commented):
could you add a test?

njzjz (approved):

PR #639, opened 2024-04-16
CI: setup uv

njzjz:


njzjz:
uv only supports Python 3.8+...

wanghan-iapcm:
the dpdata CI is not intensive, so shall we keep pip as is?

njzjz:
I leave this PR as a draft until we want to deprecate Python 3.7.

njzjz:
A bit strange: while `pip install uv` requires Python 3.8, installing uv binary directly (it's not written in Python) doesn't have any Python requirements...

njzjz:
This PR saves about 1 minute

PR #643, opened 2024-04-18
docs: setup uv for readthedocs

njzjz:


njzjz:
Need to take a look...

Issue #642, opened 2024-04-18
[Feature Request] 


ZLI-afk:
### Summary

Add a controller to switch on/off registered post function `rot_lower_triangular` when load system from `POSCAR`

### Detailed Description

As `rot_lower_triangular` function may re-orient structure under global coordinates that omit the alignment information that may be important for structure manupulation

### Further Information, Files, and Links

please refer to original discussion [page](https://github.com/deepmodeling/dpdata/discussions/641)

PR #644, opened 2024-04-19
Add `post_func_skip_list` to control execution of post functions

ZLI-afk:
In respond to required function associated with recent discussion topic #641 and #642

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -183,7 +183,9 @@ class System(MSONable):
             return
         if file_name is None:
             return
-        self.post_func_skip_list = post_func_skip_list if post_func_skip_list is not None else []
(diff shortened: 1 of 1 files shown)


dpdata/system.py
```diff
@@ -229,7 +229,12 @@ def from_fmt_obj(self, fmtobj, file_name, **kwargs):
                 self.data = {**self.data, **data}
                 self.check_data()
             if hasattr(fmtobj.from_system, "post_func"):
+                post_func_skip_list = kwargs.get("post_func_skip_list", [])
```
njzjz, line 232: The argument should be written explicitly, otherwise others won't realize this argument.

njzjz (changes requested):

Issue #526, opened 2023-08-16
slow to import dpdata


link89:
I find that it would takes up to 30s to run just `import dpdata`  on the login node of HPC.  I did some investigation myself and find that it would load almost all the modules into memory when importing the `dpdata` module. 

Here is the relative code:
https://github.com/deepmodeling/dpdata/blob/master/dpdata/__init__.py#L21-L30

Is it possible to just remove the `__all__`  expression? I think it is for users who like to use `from dpdata import *`,  which is not a good practices IMO.

njzjz:
![image](https://github.com/deepmodeling/dpdata/assets/9496702/8e8729de-e2ea-4cc6-af82-1a298818438f)

```
import time: self [us] | cumulative | imported package
import time:       151 |        151 |   _io
import time:        51 |         51 |   marshal
import time:       231 |        231 |   posix
import time:       371 |        802 | _frozen_importlib_external
import time:        56 |         56 |   time
import time:       171 |        227 | zipimport
import time:        41 |         41 |     _codecs
import time:       328 |        369 |   codecs
import time:       249 |        249 |   encodings.aliases
import time:       360 |        977 | encodings
import time:       123 |        123 | encodings.utf_8
import time:       115 |        115 | _signal
import time:       145 |        145 | encodings.latin_1
import time:        28 |         28 |     _abc
import time:       150 |        177 |   abc
import time:       156 |        332 | io
import time:        93 |         93 |       _stat
import time:       166 |        258 |     stat
import time:       772 |        772 |     _collections_abc
import time:        94 |         94 |       genericpath
import time:       149 |        242 |     posixpath
import time:       362 |       1633 |   os
import time:       111 |        111 |   _sitebuiltins
import time:        49 |         49 |     _locale
import time:        89 |        137 |   _bootlocale
import time:       249 |        249 |         types
import time:       182 |        182 |         warnings
import time:       104 |        533 |       importlib
import time:       121 |        654 |     importlib.machinery
import time:       156 |        156 |               _heapq
import time:       161 |        316 |             heapq
import time:       634 |        634 |             itertools
import time:       127 |        127 |             keyword
import time:        62 |         62 |               _operator
import time:       268 |        330 |             operator
import time:       147 |        147 |             reprlib
import time:        61 |         61 |             _collections
import time:       887 |       2500 |           collections
import time:       187 |        187 |           collections.abc
import time:        47 |         47 |               _functools
import time:       460 |        507 |             functools
[...]

PR #653, opened 2024-05-11
benchmark performance of import

njzjz:
Do a benchmark before working on #526

## Summary by CodeRabbit

- **New Features**
	- Introduced benchmark tests for import functionality and command line interface operations.
- **Tests**
	- Added new benchmark tests to assess performance improvements.
- **Chores**
	- Integrated `pytest` and `pytest-codspeed` into the project for enhanced testing capabilities.


wanghan-iapcm (approved):

adriencaccia:
Hey @njzjz, I am a co-founder @CodSpeedHQ, glad to see you trying out our tool!

You use the CodSpeed action inside a matrix, thus running the action 3 times with different Python versions.
However, this will not work properly as CodSpeed is built to run only once per PR. There might be instances where the job is picked up correctly, and not the same Python version will be picked up each time.

I recommend that you create a separate workflow file or job with a single Python version and single use of the CodSpeed action. I recommend using Python 3.12, so that you can benefit from the automated flamegraphs in the CodSpeed UI 😉

njzjz:
> I recommend that you create a separate workflow file or job with a single Python version and single use of the CodSpeed action. I recommend using Python 3.12, so that you can benefit from the automated flamegraphs in the CodSpeed UI 😉

Thanks for your friendly reminder!

PR #660, opened 2024-05-17
improve: ase try to get virials from different sources

link89:
## Summary by CodeRabbit

- **Enhancements**
  - Improved the method for obtaining virials in the system by checking multiple sources before calculation, ensuring more accurate data handling.
- **Chores**
  - Added a new optional dependency to support documentation building.

### DIFF
--- a/dpdata/plugins/ase.py
+++ b/dpdata/plugins/ase.py
@@ -63,6 +63,8 @@ class ASEStructureFormat(Format):
         """Convert ase.Atoms to a LabeledSystem. Energies and forces
         are calculated by the calculator.
 
+        Note that this method will try to load virials from either virial field or converted from stress tensor.
+
         Parameters
         ----------
         atoms : ase.Atoms
@@ -94,13 +96,19 @@ class ASEStructureFormat(Format):
             "energies": np.array([energies]),
             "forces": np.array([forces]),
         }
-        try:
-            stress = atoms.get_stress(voigt=False)
-        except PropertyNotImplementedError:
-            pass
-        else:
-            virials = np.array([-atoms.get_volume() * stress])
-            info_dict["virials"] = virials
+
+        # try to get virials from different sources
+        virials = atoms.info.get("virial")
+        if virials is None:
+            try:
+                stress = atoms.get_stress(voigt=False)
+            except PropertyNotImplementedError:
+                pass
+            else:
+                virials = -atoms.get_volume() * stress
+        if virials is not None:
+            info_dict["virials"] = np.array([virials])
+
         return info_dict
 
     def from_multi_systems(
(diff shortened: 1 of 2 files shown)


dpdata/plugins/ase.py
```diff
@@ -93,13 +93,21 @@ def from_labeled_system(self, atoms: ase.Atoms, **kwargs) -> dict:
...
-            stress = atoms.get_stress(False)
-        except PropertyNotImplementedError:
-            pass
-        else:
-            virials = np.array([-atoms.get_volume() * stress])
+
+        # try to get virials from different sources
+        virials = atoms.info.get("virial")
+        if virials is None:
+            virials = atoms.info.get("virials")
```
njzjz, line 100: It's unclear why it has two different keys.

link89: It's a fallback strategy as users may use either `virial` or `virials`.

njzjz: Documentation is necessary if users are expected to do something

link89: I try to search in the code base but find no place to add comment for this. 

I don't think we need add extra documents for this as `dpdata` is supposed to be able to find `virial` for user automatically. My patch doesn't introduce any compatibility issue, it just make the ase plugin more robust to find `virial`.

njzjz: It finds data from non-standard keys, which needs documentation to avoid unexpected behaviors.

link89: I just add a comment in docstring.

njzjz: It's not rendered correctly.
image

dpdata/plugins/ase.py
```diff
@@ -62,6 +62,11 @@ def from_labeled_system(self, atoms: ase.Atoms, **kwargs) -> dict:
         """Convert ase.Atoms to a LabeledSystem. Energies and forces
         are calculated by the calculator.
 
+        Note that this method will try to load virials from the following sources:
+        - atoms.info['virial']
+        - atoms.info['virials']
+        - converted from stress tensor
```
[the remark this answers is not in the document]
njzjz: Yes, a blank line needs to be added.

njzjz:
Please let all checks pass.

link89:
Resolved. I think the code coverage is a false positive.

pyproject.toml
```diff
@@ -59,6 +59,7 @@ docs = [
     'rdkit',
     'jupyterlite-sphinx',
     'jupyterlite-xeus',
+    'm2r>=0.3.1',
```
njzjz: What's the purpose of this change?

link89: fix failing doc build in the previous run: https://app.readthedocs.org/api/v2/build/28872644.txt

njzjz: The log shows m2r2 other than m2r. M2r2 was cleaned in #701. Please merge from the base branch.

njzjz (approved):

PR #669, opened 2024-06-06
abacus: fix bug in collecting force

pxlxingliang:
## Summary by CodeRabbit

- **Bug Fixes**
  - Modified force data handling in computational simulations to return `None` instead of an empty list when no forces are present.

- **Tests**
  - Added new test scenarios to validate force data handling in labeled system objects for computational simulations.

### DIFF
--- a/dpdata/abacus/scf.py
+++ b/dpdata/abacus/scf.py
@@ -190,7 +190,7 @@ def collect_force(outlines):
 def get_force(outlines, natoms):
     force = collect_force(outlines)
     if len(force) == 0:
-        return [[]]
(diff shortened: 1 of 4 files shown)


njzjz:
I am wondering if it works, considering `System` will check the shape of the data.

pxlxingliang:
> I am wondering if it works, considering `System` will check the shape of the data.

In current dpdata, it does work. It seems that the `System` does not check the dimension of force.

Is there a specification for how to store force in a system without force?

njzjz (approved):

PR #744, opened 2024-10-26
Add support for converting move flags between Abacus and VASP

Angel-Jia:
## Release Notes

- **New Features**
	- Enhanced atomic movement handling with the addition of a `move` parameter across multiple functions, enabling more detailed tracking of atom movements.
	- Introduced selective dynamics support in VASP POSCAR data handling, allowing for better control over atomic movement flags.
	- Added structured representation for cubic boron nitride crystal in new test files.

- **Bug Fixes**
	- Updated atomic positions and magnetic properties in test structures to reflect accurate simulation data.

- **Tests**
	- Added new test methods to validate the handling of movement flags and ensure the integrity of output structures across various scenarios.


Angel-Jia:
Currently, only input files POSCAR for VASP and STRU for Abacus are supported. The move flags in their output files have not been handled yet.

wanghan-iapcm (changes requested):
The newly defined data type `"move"` should be registered. for example 
https://github.com/deepmodeling/dpdata/blob/7853b7b526a5ffbd2b8a08d3a1f3ffa06c228d52/dpdata/plugins/abacus.py#L49

otherwise, it will not handled by dpdata, which may be tested by load and dump the same poscar file.

dpdata/plugins/abacus.py
```diff
@@ -65,6 +65,15 @@ def register_mag_data(data):
...
         dpdata.LabeledSystem.register_data_type(dt)
+    if "move" in data:
+        dt = DataType(
+            "move",
+            np.ndarray,
+            (Axis.NFRAMES, Axis.NATOMS, 3),
+            required=False,
+            deepmd_name="move",
+        )
+        dpdata.System.register_data_type(dt)
```
wanghan-iapcm, line 76: not a good idea to register "move" in the function `register_mag_data`, which registers the magnatic related variable and labels.

wanghan-iapcm (changes requested):

wanghan-iapcm (approved):

PR #800, opened 2025-03-06
abacus/stru: return spins only when atomic magnetic moment is specified

pxlxingliang:
The old version will induce a bug that dpdata will always store the spin information, and when dpdata save it to lammps format, the spin information will also be written, which is invalid for a non-spin lammps job.
Now, the spin information is saved to dpdata only when the magnetic moment of one atom is specified, which is necessary for a spin job.

## Summary by CodeRabbit

- **New Features**
  - Spin data is now processed to include only explicitly defined moments, ensuring accurate output.
- **Documentation**
  - User-facing documentation has been updated to clarify when spin information appears.
- **Tests**
  - Test entries for atomic positions have been streamlined by removing redundant markers.


wanghan-iapcm (approved):

pxlxingliang:
Hi @wanghan-iapcm @njzjz , the bug fixed by this commit can cause the failure of using DPGEN+ABACUS. Could you please release a new version?

PR #803, opened 2025-03-11
update coords after shift_orig_zero

zrzrv5:
Hi,

I think this part of the code was attempting to modify local variables (`ii`) rather than the actual array elements in `self.data["coords"]`. 

Before:
```
@post_funcs.register("shift_orig_zero")
    def _shift_orig_zero(self):
        for ff in self.data["coords"]:
             for ii in ff:
                 ii = ii - self.data["orig"]
                 ...
```
After:
```
...
self.data["coords"] = self.data["coords"] - self.data["orig"]
```

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the efficiency of coordinate adjustments for improved performance on larger datasets.

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -710,9 +710,7 @@ class System:
 
     @post_funcs.register("shift_orig_zero")
     def _shift_orig_zero(self):
-        for ff in self.data["coords"]:
-            for ii in ff:
-                ii = ii - self.data["orig"]
+        self.data["coords"] = self.data["coords"] - self.data["orig"]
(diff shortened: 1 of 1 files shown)


njzjz:
... If this method never works at all for six years, I am wondering why not remove it.

wanghan-iapcm (approved):

wanghan-iapcm:
> ... If this method never works at all for six years, I am wondering why not remove it.

there are marginal cases where the origin is not [0,0,0]

PR #834, opened 2025-06-16
Add specified elements for perturbation

tang070205:
Adding the `elem_pert_list` parameter can perturb the elements of a specified element, which I personally think may be necessary.


wanghan-iapcm:
please use https://github.com/deepmodeling/dpdata/tree/master/tests/test*py and https://github.com/deepmodeling/dpdata/tree/master/tests/*/test*py as examples and write your own unittests for testing your implementation.

PR #841, opened 2025-07-03
ABACUS: rotate to fit right hand rule when reading STRU

pxlxingliang:
## Summary by CodeRabbit

* **New Features**
  * Ensured that cell vectors and atomic coordinates now consistently follow the right-hand rule, improving the reliability of structural data representation.


wanghan-iapcm:
> I don't think there are any checks in other formats to ensure the right-hand rule.

the stru structure is internally interpreted to adhere to the right hand rule in abacus. this pr is to make sure that  the cell in dpdata aligns with that used in the abacus.

PR #842, opened 2025-07-03
lammps: rotate cell to lower_triangle before writing to lammps

pxlxingliang:


njzjz (changes requested):
As discussed with @wanghan-iapcm ,  #841 should also be applied to LAMMPS, since we cannot promise dpdata data has obeyed any rule.

wanghan-iapcm:
> > As discussed with @wanghan-iapcm , #841 should also be applied to LAMMPS, since we cannot promise dpdata data has obeyed any rule.
> 
> Hi, @njzjz, the newly added function will check the diagonal terms of the lower triangular matrix to ensure that the diagonal is not negative, which can keep the determinant value of the cell not less than 0, which can ensure that the cell meets the right-hand rule.

dpdata.lammps is expected to handle the  left-handed cells, which may be derived from plugins that does not obey the right-hand rule.

pxlxingliang:
> > convert the diagonal terms to positive values to ensure that the cell satisfies the right-hand rule
> 
> Why are these two related?
> 
> Besides, the doc for the method doesn't mention it.

For a lower triangular matrix, if the diagonal terms have positive values, then the determinant of the matrix is positive, which satisfies the right-hand rule.

njzjz (approved):

PR #852, opened 2025-07-29
Fix incorrect spin-element mapping in make_unlabeled_stru() for ABACUS STRU generation

aboys-cb:
This patch resolves a bug in dpdata/abacus/stru.py where the spin values were misaligned with their corresponding elements when generating an unlabeled STRU file.
Root cause:
In make_unlabeled_stru(), the code correctly identifies the global atom index (iatomtype) for each element type, but then mistakenly uses a separate running counter (natom_tot) to index every property except atomic coordinates. As a result, spins (and any other per-atom attributes) are pulled from the wrong positions in the input array, breaking the element–spin correspondence.
Fix:
Replace all usages of natom_tot with iatomtype when accessing per-atom data (spins, velocities, etc.). Coordinates were already using the correct index; the change makes the remaining properties consistent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy in assigning per-atom properties when generating atomic position data, ensuring correct values are used for each atom.

### DIFF
--- a/dpdata/abacus/stru.py
+++ b/dpdata/abacus/stru.py
@@ -731,7 +731,9 @@ def make_unlabeled_stru(
         out += "0.0\n"
         out += str(data["atom_numbs"][iele]) + "\n"
         for iatom in range(data["atom_numbs"][iele]):
-            iatomtype = np.nonzero(data["atom_types"] == iele)[0][iatom] # it is the atom index
+            iatomtype = np.nonzero(data["atom_types"] == iele)[0][
+                iatom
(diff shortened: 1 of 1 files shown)


pxlxingliang:
The original version will indeed have problems, when the atomic order is not strictly arranged according to element types.

pxlxingliang (approved):

njzjz:
> The original version will indeed have problems, when the atomic order is not strictly arranged according to element types.

@pxlxingliang Could we add a test for this case?

PR #856, opened 2025-08-01
fix: skip datatype registration warning for duplicate types

caic99:
Currently dpdata throws a warning if an existed data type is registered again, even if with the same definition. This PR adds a check to skip if the registered data types are identicle.

## Summary by CodeRabbit

* **New Features**
  * Improved object comparison and display for data types, allowing equality checks and clearer string representations.

* **Bug Fixes**
  * Reduced unnecessary warnings when registering identical data types; warnings are now only shown if duplicate registrations differ.

* **Tests**
  * Added tests for data type comparison, representation, and registration behavior.
  * Removed outdated test for duplicate data type registration.

### DIFF
--- a/dpdata/system.py
+++ b/dpdata/system.py
@@ -1102,12 +1102,11 @@ class System:
         all_dtypes = cls.DTYPES + tuple(data_type)
         dtypes_dict = {}
         for dt in all_dtypes:
-            if dt.name in dtypes_dict:
-                if dt != dtypes_dict[dt.name]:
-                    warnings.warn(
-                        f"Data type {dt.name} is registered twice with different definitions; only the newly registered one will be used.",
-                        UserWarning,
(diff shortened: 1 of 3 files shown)


dpdata/system.py
```diff
@@ -1103,10 +1103,11 @@ def register_data_type(cls, *data_type: DataType):
         dtypes_dict = {}
         for dt in all_dtypes:
             if dt.name in dtypes_dict:
-                warnings.warn(
-                    f"Data type {dt.name} is registered twice; only the newly registered one will be used.",
-                    UserWarning,
-                )
+                if dt != dtypes_dict[dt.name]:
```
njzjz: Lines 1105 and 1106 can be merged to one line.

njzjz (approved):

PR #875, opened 2025-08-27
Latest support for abacus scf output

kluophysics:
Abacus updates its output format, so these keywords were not right. I have updated them to support both eariler ones and latest one by removing the unit part. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness when reading ABACUS simulation outputs by accepting more variations of energy, force, and stress headers.
  * More permissive header matching reduces failures from minor log-format differences.
  * Preserves existing convergence and extraction behavior while improving tolerance for alternate log phrasing.
  * Reduces missed or incomplete parses, increasing reliability across ABACUS log variants.

### DIFF
--- a/dpdata/abacus/scf.py
+++ b/dpdata/abacus/scf.py
@@ -45,10 +45,10 @@ def get_path_out(fname, inlines):
 def get_energy(outlines):
     Etot = None
     for line in reversed(outlines):
-        if "final etot is" in line:
+        if "final etot is" in line: # for LTS
             Etot = float(line.split()[-2])  # in eV
-        if "ETOT" in line:
-            Etot = float(line.split()[1])  # in eV
+        elif "TOTAL ENERGY" in line: # for develop
+            Etot = float(line.split()[-2])  # in eV
             return Etot, True
         elif "convergence has NOT been achieved!" in line:
             return Etot, False
--- a/dpdata/abacus/scf.py
+++ b/dpdata/abacus/scf.py
@@ -45,9 +45,9 @@ def get_path_out(fname, inlines):
 def get_energy(outlines):
     Etot = None
     for line in reversed(outlines):
-        if "final etot is" in line: # for LTS
+        if "final etot is" in line:  # for LTS
             Etot = float(line.split()[-2])  # in eV
-        elif "TOTAL ENERGY" in line: # for develop
+        elif "TOTAL ENERGY" in line:  # for develop
             Etot = float(line.split()[-2])  # in eV
             return Etot, True
         elif "convergence has NOT been achieved!" in line:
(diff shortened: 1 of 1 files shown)


dpdata/abacus/scf.py
```diff
@@ -45,8 +45,8 @@ def get_path_out(fname, inlines):
 def get_energy(outlines):
     Etot = None
     for line in reversed(outlines):
-        if "final etot is" in line:
-            Etot = float(line.split()[-2])  # in eV
+        if "ETOT" in line:
+            Etot = float(line.split()[1])  # in eV
```
pxlxingliang: This modification will cause that dpdata can not catch the converged energy for previous version, based on the lasted ABACUS (
https://github.com/deepmodeling/abacus-develop/blob/37bdf71826c2a65729a8e7338c5be26257102b47/source/source_io/output_log.cpp#L18-L23), I suggest that add another case with key "TOTAL ENERGY" for new version , like:
```
if "final etot is" in line:
   ...
elif "TOTAL ENERGY" line line: # for new version 
   ...
elif ...
```

pxlxingliang: We should not use ETOT as the key-word here. I guess that you want to catch the energy in line ` !FINAL_ETOT_IS xxx eV`, which is always outputted whether the SCF is converged or not. 
The second returned value of this function `get_energy()` is used to tell if the SCF is converged or not. So we can not directly return (Etot, True)  when "ETOT" key-word is found. While, key-word "final etot is" (or "TOTAL ENERGY" in new version) is outputted only when the SCF is converged. In dpdata, we avoid to save the energy of a not converged SCF job.

kluophysics: Okay, fixed.

dpdata/abacus/scf.py
```diff
@@ -45,7 +45,9 @@ def get_path_out(fname, inlines):
 def get_energy(outlines):
     Etot = None
     for line in reversed(outlines):
-        if "final etot is" in line:
+        if "final etot is" in line:  # for LTS
+            Etot = float(line.split()[-2])  # in eV
+        elif "TOTAL ENERGY" in line:  # for develop
```
pxlxingliang, line 51: Should return (Etot True) for first case of if:
```
        if "final etot is" in line:  # for LTS
            Etot = float(line.split()[-2])  # in eV
            return Etot, True
        elif "TOTAL ENERGY" in line:  # for develop
            Etot = float(line.split()[-2])  # in eV
            return Etot, True
...
```

wanghan-iapcm:
> One additional question: What's better for the energy in training dataset? the total energy from ETOT, or the energy(sigma->0) from E_KohnSham?

Both are OK if you were only train your model with energy.
When you train your model with force, then ETOT as the force is the negative gradient of the ETOT.

PR #896, opened 2025-09-16
Add gaussian/fchk format support

Romarin87:
Added reading of fchk format files and support for saving hessian.npy

## Summary by CodeRabbit

* **New Features**
  * Added import support for Gaussian formatted checkpoint (.fchk) with automatic unit conversions for coordinates, energies, forces, and Hessians; Hessian data type is now supported and available; new Gaussian FCHK format handler added; non-periodic systems handled by default.

* **Tests**
  * Added FCHK/LOG fixtures and unit tests validating presence, shapes, and cross-file consistency of energies, forces, coordinates, atom info, and Hessians.


dpdata/plugins/gaussian.py
```diff
@@ -15,6 +19,19 @@
...
+def register_hessian_data(data):
+    if "hessian" in data:
+        dt = DataType(
+            "hessian",
+            np.ndarray,
+            (Axis.NFRAMES, -1, -1),
+            required=False,
+            deepmd_name="hessian",
+        )
+        dpdata.System.register_data_type(dt)
```
njzjz: I don't think `System` needs Hessian which is considered as labeled data.

dpdata/plugins/gaussian.py
```diff
@@ -15,6 +19,19 @@
     from dpdata.utils import FileType
 
 
+def register_hessian_data(data):
+    if "hessian" in data:
+        dt = DataType(
+            "hessian",
+            np.ndarray,
+            (Axis.NFRAMES, -1, -1),
```
njzjz: Why using -1? The shape should be `(Axis.NFRAMES, Axis.NATOMS, 3, Axis.NATOMS, 3)`.

njzjz (approved):

PR #945, opened 2026-02-27
chore(lmdb): allow missing LMDB imports

njzjz:
## Summary by CodeRabbit

- **Bug Fixes**
  - Core package functionality, including POSCAR loading, remains available when optional LMDB dependencies are not installed.
  - LMDB operations now provide a clear installation-related error when required dependencies are unavailable.

- **Documentation**
  - Improved the interactive documentation experience with updated notebook embedding and content loading.
  - Simplified the online tutorial notebook and improved file resolution for different environments.

- **Tests**
  - Added coverage for dependency-independent imports and expected LMDB behavior.

### DIFF
--- a/.github/workflows/test_import.yml
+++ b/.github/workflows/test_import.yml
@@ -33,3 +33,23 @@ jobs:
         assert loaded.formula == system.formula
         assert loaded.get_nframes() == system.get_nframes()
         PY
+    - name: Test core import without LMDB dependencies
+      run: |
+        python -m pip uninstall -y lmdb msgpack
+        python - <<'PY'
+        import dpdata
+
+        system = dpdata.System(
+            "tests/poscars/POSCAR.h2o.md", fmt="vasp/poscar"
+        )
+        assert system.get_nframes() == 1
+
+        try:
+            dpdata.System("missing.lmdb", fmt="lmdb")
+        except ModuleNotFoundError as exc:
+            assert "lmdb" in str(exc)
+        else:
+            raise AssertionError(
+                "LMDB format should require its backend dependencies"
+            )
+        PY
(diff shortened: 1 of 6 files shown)


njzjz:
Closing: the win this PR was after has already landed via #1012.

This branch no longer rebases — #1012 moved `dpdata/lmdb/format.py` to `dpdata/formats/lmdb/format.py` and rewrote it, so re-applying would mean redoing the change against a different module rather than replaying the diff. Before doing that I measured what is actually left to gain, on master `c77e038`, Python 3.13.9, median of 5 runs of `python -X importtime -c "import dpdata"`:

```
import dpdata  total          : 234.0 ms
  rdkit subtree               :  51.4 ms
  parmed subtree              :  50.4 ms
  lmdb + msgpack subtree      :   1.4 ms
import msgpack_numpy (alone)  :  43.9 ms
```

This PR deferred three imports — `lmdb`, `msgpack`, and `msgpack_numpy` — plus an `__init__` whose job was to postpone `m.patch()`. The expensive one was `msgpack_numpy` at ~44 ms, and #1012 removed it outright: `grep -rn msgpack_numpy dpdata/` returns nothing, and `pyproject.toml` now lists only `lmdb>=2.0.0` and `msgpack`. The `__init__` added here has nothing left to patch.

What remains is `lmdb` (0.6 ms) + `msgpack` (1.4 ms), about 0.6% of import time, against pushing every `lmdb.*` / `msgpack.*` reference in a freshly rewritten ~1000-line module into function bodies, including four module-level annotations (`_READ_ENV_CACHE`, `_read_metadata`, `_open_read_env`, `_open_publish_guard`) that would have to move under `TYPE_CHECKING`. Not worth the churn on code that was just rewritten.

The real remaining import cost is elsewhere: `dpdata/plugins/rdkit.py:3` imports `dpdata.formats.rdkit.utils` eagerly, and parmed comes in through the AMBER path — together ~102 ms, 44% of `import dpdata`. Worth a separate PR if import time matters.

PR #979, opened 2026-06-19
fix(serialization): remove monty dependency

njzjz:
## Summary
- replace monty serialization usage with an internal dpdata.serialization module
- keep loading existing monty-style JSON numpy arrays and as_dict/from_dict objects
- remove monty from project/docs dependency declarations and update JSON round-trip tests

## Tests
- ruff check dpdata/ tests/test_json.py tests/test_to_pymatgen_entry.py
- cd tests && python -m unittest test_json.py test_to_pymatgen_entry.py

## Summary by CodeRabbit

* **New Features**
  * Added native dump/load helpers for JSON, YAML, and msgpack with automatic format detection and transparent gzip/bzip2 compression.
  * Serialization now round-trips common types (e.g., complex arrays, datetimes, UUIDs, enums, and paths).
  * System dump/load and dict reconstruction now use the new serializer for consistent decoding.
* **Chores**
  * Removed `monty` from runtime dependencies and updated lint rules accordingly.
* **Documentation**
  * Updated installation/troubleshooting docs and intersphinx/docs environment config to remove `monty`.
* **Tests**
  * Expanded JSON/YAML serialization coverage with temp-file round-trips, nested objects, and timezone-aware datetimes.


dpdata/serialization.py
```diff
@@ -0,0 +1,226 @@
...
+    try:
+        yaml = importlib.import_module("yaml")
+
+        return getattr(yaml, "safe_load")(fp, *args, **kwargs)
+    except ModuleNotFoundError:
+        try:
+            ruamel_yaml = importlib.import_module("ruamel.yaml")
+        except ModuleNotFoundError as e:
+            raise RuntimeError(
+                "Loading YAML files requires PyYAML or ruamel.yaml."
```
wanghan-iapcm, line 86: ### Code review

Found 1 issue:

1. Dropping `monty` removes the only YAML backend, so the documented `.yaml` dump/load path now raises on a clean install.

`monty` declared `ruamel.yaml` as a hard dependency, so `System.dump("x.yaml")` / `System.load("x.yaml")` worked out of the box. This PR removes `monty` from `dependencies` in `pyproject.toml` without adding PyYAML or ruamel.yaml, while `System.dump`/`System.load` still advertise `.yaml`. With neither backend importable, `_yaml_load`/`_yaml_dump` raise `RuntimeError` here, a regression for a documented format. Consider adding a YAML backend to `dependencies` (or an optional extra) and/or noting the requirement in the `dump`/`load` docstrings.

dpdata/serialization.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _detect_format(filename: str | Path, fmt: str | None = None) -> str:
+    if fmt is not None:
+        return fmt
+    basename = Path(filename).name.lower()
+    if ".mpk" in basename:
+        return "mpk"
+    if ".yaml" in basename or ".yml" in basename:
+        return "yaml"
+    return "json"
```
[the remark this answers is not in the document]
njzjz: Fixed in commit b1e32a5. Format detection now examines the final meaningful suffix after removing a trailing compression suffix, so filenames ending in .yaml.json and .mpk.backup correctly default to JSON. Added focused regression coverage.

Coding agent: Codex
Codex version: codex-cli 0.144.4
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/serialization.py
```diff
@@ -0,0 +1,229 @@
...
+            if cls is not None:
+                data = {k: v for k, v in obj.items() if not k.startswith("@")}
+                if hasattr(cls, "from_dict"):
+                    return cls.from_dict(data)
+                if isinstance(cls, type) and issubclass(cls, Enum):
+                    return cls(process_decoded(data["value"]))
+        return {process_decoded(k): process_decoded(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [process_decoded(v) for v in obj]
+    return obj
```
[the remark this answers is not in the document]
njzjz: Fixed in commit b1e32a5. Values are now recursively decoded before being passed to a custom class from_dict method, with regression coverage for nested NumPy arrays and datetimes.

Coding agent: Codex
Codex version: codex-cli 0.144.4
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/serialization.py
njzjz: The YAML backend regression remains addressed: PyYAML is declared as a core dependency and the YAML dump/load regression test passes on the current head.\n\nCoding agent: Codex\nCodex version: codex-cli 0.144.4\nModel: gpt-5.6-sol\nReasoning effort: xhigh

dpdata/system.py
```diff
@@ -333,7 +333,7 @@ def __add__(self, others):
 
     def dump(self, filename: str, indent: int = 4):
         """Dump .json or .yaml file."""
```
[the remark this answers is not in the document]
njzjz: Addressed in commit `adbdaf5`. The `System.dump()` docstring now documents JSON, YAML, and MessagePack support.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

tests/test_json.py
```diff
@@ -26,5 +54,73 @@ def setUp(self):
...
+        self.filename = os.path.join(self.tmpdir.name, "h2o.md.json")
+        self.system_1.dump(self.filename)
+        self.system_2 = dpdata.LabeledSystem.load(self.filename)
+        self.places = 6
+        self.e_places = 6
+        self.f_places = 6
+        self.v_places = 4
+
+    def tearDown(self):
+        self.tmpdir.cleanup()
```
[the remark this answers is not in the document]
njzjz: Addressed in commit `adbdaf5`. The temporary directory cleanup is registered immediately with `addCleanup()`, and the former `tearDown()` cleanup was removed so setup failures cannot leak the directory.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/system.py
```diff
@@ -380,19 +380,17 @@ def map_atom_types(
     @staticmethod
     def load(filename: str):
         """Rebuild System obj. from .json or .yaml file."""
```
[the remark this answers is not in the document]
njzjz: Addressed in commit `adbdaf5`. The `System.load()` docstring now documents JSON, YAML, and MessagePack support.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

tests/test_json.py
```diff
@@ -1,10 +1,34 @@
...
+class NestedSerializable:
+    """Test helper that records values passed through ``from_dict``."""
+
+    def __init__(self, value):
+        self.value = value
+
+    @classmethod
+    def from_dict(cls, data):
+        """Construct the helper from decoded serialized data."""
+        return cls(data["value"])
```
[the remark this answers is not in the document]
njzjz: Addressed in commit `278fb69`. `NestedSerializable` now implements `as_dict()`, so the regression test genuinely exercises recursive serialization and decoding of the nested NumPy array and datetime. The focused serialization test suite passes.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/serialization.py
```diff
@@ -0,0 +1,238 @@
...
+    if isinstance(obj, np.ndarray):
+        return _encode_ndarray(obj)
+    if isinstance(obj, np.generic):
+        return obj.item()
+    if isinstance(obj, datetime.datetime):
+        return {
+            "@module": "datetime",
+            "@class": "datetime",
+            "string": str(obj),
+        }
```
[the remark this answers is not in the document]
njzjz: Addressed in commit `adbdaf5`. Datetimes are now serialized with `isoformat()`, and the tests assert the exact serialized representation and timezone-aware round trip.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

njzjz:
Closed the `codecov/patch` gap in acaef94.

`dpdata/serialization.py` was at 64%. The uncovered lines were the paths no existing test reaches: msgpack dump/load, the gzip/bz2 streams, `_open_binary` entirely, the ruamel fallback and its `RuntimeError`, complex-array encode/decode, UUID/Path/Enum encoding, the datetime `strptime` fallbacks, the unknown-`@module` branch, and `TypeError: Invalid format`.

`tests/test_serialization.py` adds 35 cases and takes the module to **100%**.

Two notes on how the harder branches are reached:

- The optional-dependency errors are triggered by patching `importlib.import_module` and `sys.modules` rather than by uninstalling anything, so PyYAML-missing, ruamel-missing, and msgpack-missing each assert their own message with the test environment untouched.
- The `strptime` fallbacks needed inputs that `fromisoformat` actually rejects. On Python 3.11+ a space separator is accepted, so `"2020-01-02 03:04:05.678900"` never reaches them; unpadded fields do, hence `"2020-1-2 3:4:5.678900"`. The `split("+")[0]` line is covered by `"2020-1-2 3:4:5+00:00"`, which also documents that the fallback silently drops an offset it cannot parse.

`python -m unittest discover` — 2201 passed, 43 skipped.

pyproject.toml
```diff
@@ -20,9 +20,9 @@ classifiers = [
 ]
 dependencies = [
     'numpy>=1.14.3',
-    'monty',
     'scipy',
     'h5py',
+    'PyYAML',
```
wanghan-iapcm, line 25: This line is the actual fix, but nothing in the suite guards it.

`_yaml_dump` / `_yaml_load` are byte-identical before and after this PR, so the only pre-fix/post-fix difference is this declaration -- and a round-trip test running inside an already-provisioned interpreter cannot observe a metadata declaration. CI installs `.[test,amber,ase,pymatgen]`, and `pymatgen -> monty -> ruamel.yaml`, so a YAML backend is always importable; I confirmed the YAML tests pass on the pre-fix commit and fail only when both backends are blocked with an import hook. So the added "regression test" passes on the broken code too.

A cheap genuine guard already almost exists: `.github/workflows/test_import.yml` does a core-only `uv pip install --system .` and then just runs `python -c 'import dpdata'`. Extending that to an actual `.yaml` dump/load would fail pre-fix and pass post-fix.

dpdata/system.py
```diff
@@ -332,8 +332,8 @@ def __add__(self, others):
         return self.__class__.from_dict({"data": self_copy.data})
 
     def dump(self, filename: str, indent: int = 4):
-        """Dump .json or .yaml file."""
-        from monty.serialization import dumpfn
+        """Dump a JSON, YAML, or MessagePack file."""
```
wanghan-iapcm, line 335: Two things on this line.

`indent` is forwarded unconditionally, but `_detect_format` routes `.mpk` into `msgpack.dump(obj, fp, **kwargs)`, so `System.dump("x.mpk")` raises `TypeError: Packer.__init__() got an unexpected keyword argument 'indent'` -- while this PR newly changes the docstring from "Dump .json or .yaml file" to advertise MessagePack. `dumpfn(s.as_dict(), "x.mpk")` works when called directly, so gating `indent` to the text formats is enough.

Separately, please add a System-level YAML test. Both current YAML tests use plain dicts, so the `indent=4` + numpy-payload + System-reconstruction path this PR actually repairs has no coverage. A `TestYamlDumpLoad` mirroring the existing `TestJsonDumpLoad` is about 12 lines, errors 15/15 on master, and passes here -- a proper regression guard for the fix.

dpdata/serialization.py
```diff
@@ -0,0 +1,238 @@
...
+        suffixes.pop()
+    suffix = suffixes[-1] if suffixes else ""
+    if suffix == ".mpk":
+        return "mpk"
+    if suffix in {".yaml", ".yml"}:
+        return "yaml"
+    return "json"
+
+
+def _open_text(filename: str | Path, mode: str) -> TextIO:
```
wanghan-iapcm, line 31: This drops `.xz` / `.lzma`, which monty's `zopen` handled, and it fails silently in both directions:

- `dumpfn(obj, "x.json.xz")` now writes plain JSON under an `.xz` name -- `file` reports "JSON text data" where master reports "XZ compressed data, checksum CRC64". No error, no warning.
- Reading a `.xz` file written by an older dpdata raises `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfd in position 0`.

Silent wrong output is worse than a hard failure here. Either add the `lzma` branch alongside `.gz`/`.z`/`.bz2`, or raise on an unrecognized compression suffix so the loss cannot go unnoticed.

wanghan-iapcm (changes requested):
Correcting my earlier review first: I claimed YAML dump/load "worked out of the box" under monty. It did not. `monty.serialization.dumpfn` sets `cls=MontyEncoder` only in the json branch, so on master `System.dump("x.yaml")` raises `TypeError: YAML.dump() got an unexpected keyword argument 'indent'` (and without `indent`, `RepresenterError: cannot represent an object: np.int64(2)`), while `loadfn` returns a raw `CommentedMap` rather than a System. So the System-level YAML path never worked, and this PR genuinely fixes it.

The dependency half of my concern was real and is properly addressed: PyYAML is in the core `dependencies` array, monty is fully removed, and JSON output is byte-identical to monty's, so files round-trip in both directions with older dpdata.

Three things below I would like resolved before merge -- the silent `.xz` behavior in particular.

pyproject.toml
njzjz: Fixed in `9d522f7`. The core-only import workflow now loads a System, dumps it to YAML, reloads it, and checks its formula and frame count after `uv pip install --system .`. This exercises the declared core YAML dependency without relying on optional packages.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/system.py
njzjz: Fixed in `9d522f7`. `System.dump()` now forwards `indent` only to JSON/YAML, so MessagePack no longer receives a text-only keyword. System-level YAML and MessagePack round-trip tests now mirror the existing JSON coverage. The focused 113 tests and full 2,232-test suite pass.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/serialization.py
njzjz: Fixed in `9d522f7`. `.xz` and `.lzma` are recognized as compression suffixes and opened through the standard-library `lzma` backend for both text and binary formats. Tests verify codec round trips and the XZ magic header.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

pyproject.toml
njzjz: Follow-up `656af0a` fixes the validation job itself by passing `fmt="vasp/poscar"`; the fixture name ends in `.md`, so auto-detection selected an unsupported format in CI. The exact core-install YAML round trip now passes locally.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/system.py
njzjz: Follow-up `656af0a` expresses the MessagePack/text branch directly instead of expanding a dynamically typed kwargs dict. This preserves the behavior from `9d522f7` and clears the Pyright diagnostic reported by CI.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #1006, opened 2026-07-02
fix(psi4): honor requested input method

### LINKED ISSUE #992: [Code scan] Psi4 input writer ignores the requested method

The Psi4 input writer accepts a `method` argument, but the generated template hard-codes `WB97M-D3BJ` in the `gradient(...)` call. Passing any other method still writes `gradient("WB97M-D3BJ", ...)`.

Affected code:
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/psi4/input.py#L150-L173

The current test only verifies the default method, so this is not caught:
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/tests/test_psi4.py#L69-L96

Minimal reproducer:

```python
import dpdata

system = dpdata.LabeledSystem("tests/psi4/psi4.out", fmt="psi4/out")
system.to_psi4_inp("tmp.inp", method="hf", basis="sto-3g")
print(open("tmp.inp").read())
```

Current generated input still contains:

```text
G, wfn = gradient("WB97M-D3BJ", return_wfn=True)
```

Expected generated input should use the requested method:

```text
G, wfn = gradient("hf", return_wfn=True)
```

rathodkunj2005:
## Summary
- Use the requested `method` argument when writing Psi4 `gradient(...)` calls instead of always writing `WB97M-D3BJ`.
- Add a regression test for non-default Psi4 input methods.

Fixes #992.

## Verification
- `python -m pytest test_psi4.py -q` (from `tests/`): `17 passed in 0.24s`
- `python -m ruff check dpdata/formats/psi4/input.py tests/test_psi4.py`: `All checks passed!`
- Direct smoke check: `write_psi4_input(..., method="hf", basis="sto-3g")` emits `gradient("hf", return_wfn=True)` and no `WB97M-D3BJ`.

## Summary by CodeRabbit

* **Bug Fixes**
  * PSI4 input generation now uses the method selected by the caller instead of always falling back to a default value.
  * Added coverage to confirm generated PSI4 input reflects the requested method and basis settings.

### DIFF
--- a/dpdata/formats/psi4/input.py
+++ b/dpdata/formats/psi4/input.py
@@ -12,7 +12,7 @@ template = """molecule {{
 }}
 set basis {basis:s}
 set gradient_write on
-G, wfn = gradient("WB97M-D3BJ", return_wfn=True)
+G, wfn = gradient("{method:s}", return_wfn=True)
 wfn.energy()
 wfn.gradient().print_out()
 """
--- a/tests/test_psi4.py
+++ b/tests/test_psi4.py
@@ -95,3 +95,12 @@ class TestPsi4Input(unittest.TestCase):
             """
             ),
         )
+
+    def test_psi4_input_uses_requested_method(self):
+        system = dpdata.LabeledSystem("psi4/psi4.out", fmt="psi4/out")
+        with tempfile.NamedTemporaryFile("r") as f:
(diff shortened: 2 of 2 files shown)


njzjz (approved):
LGTM. Thanks!

PR #1008, opened 2026-07-05
fix: skip zero-force fabrication when ABACUS MD dump lacks FORCE columns

### LINKED ISSUE #1000: [Code scan] ABACUS MD can fabricate zero force labels when FORCE is absent

The ABACUS MD dump reader initializes `forces` as a zero array for all frames and atoms, fills it only when the dump header contains `FORCE`, and later always assigns `data["forces"] = force` for `LabeledSystem` output. If an ABACUS MD dump contains positions and energies but no force columns, dpdata can return valid-looking zero forces instead of omitting forces or reporting that force labels are unavailable.

Affected code:
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/abacus/md.py#L57-L75
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/abacus/md.py#L105-L124
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/abacus/md.py#L198-L214

Problematic flow:

- `calc_force` defaults to `False`.
- `forces = np.zeros([nframes_dump, total_natoms, 3])` is allocated before parsing.
- The parser only overwrites `forces[...]` inside `if calc_force:`.
- The labeled data dict always receives `data["forces"] = force`.

Expected behavior: if the dump does not contain force columns, the parser should not create zero force labels. It should either omit `forces` when returning unlabeled-compatible data, or fail clearly when the caller requests a `LabeledSystem` that requires forces for this format.

abhi-0203:
## Summary

When an ABACUS MD dump file does not contain FORCE columns (i.e., ), the parser previously initialized forces as a zero array and always assigned it to the output data dict. This fabricated misleading zero forces instead of correctly omitting the forces key.

Now  returns  for forces when  is , and  only includes  in the data dict when actual force data was parsed.

## Changes

- ****: Return  for forces when  is False; guard  and data assignment against 
- ****: Add  verifying that dumps without FORCE columns do not fabricate zero forces
- ****: New test fixture with MD dump lacking FORCE columns

## Test Plan

- [x] All existing tests pass ( — 9/9)
- [x] New test  passes
- [x] Verified that when FORCE is absent from the dump,  is not in 

Closes #1000

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of ABACUS MD outputs when force data is missing, preventing errors and ensuring frames are still parsed correctly.
  * Unconverged-frame handling now works safely even when no force values are present.

* **Tests**
  * Added coverage for MD parsing without force columns.
  * Included a new sample input and output dataset to verify coordinates, energies, and cells are still read as expected.


abhi-0203:
Addressed CodeRabbit review comments:

1. **Missing  fixture** (major): Added  with 2-frame energy data matching the MD_dump entries. The test now runs without FileNotFoundError.

2. **Redundant  check** (nitpick): Kept as-is — while the check is technically redundant when , removing it introduces a Pyright type error () because Pyright can't prove  is never None at that point. The defensive check is harmless and keeps the type checker happy.

All 10 ABACUS MD tests pass.

PR #1036, opened 2026-07-17
docs(skills): expand extxyz format documentation in dpdata-cli skill

SchrodingersCattt:
## Summary

Expand the `dpdata-cli` SKILL with detailed extended XYZ (extxyz) format documentation.

## Motivation

The extxyz format has 6 registered aliases (`extxyz`, `quip/gap/xyz`, `quip/gap/xyz_file`, `mace/xyz`, `nequip/xyz`, `gpumd/xyz`) that all point to the same `QuipGapXYZFormat` class. AI agents and users frequently do not know these are equivalent, leading to confusion. Additionally, plain `xyz` vs `extxyz` is a common source of errors.

## Changes

- Updated YAML frontmatter description to mention extxyz aliases
- Added new section with:
  - Format aliases table
  - Clear `xyz` vs `extxyz` disambiguation
  - File structure documentation
  - Recognized header key variants
  - Stress/virial conversion conventions
  - CLI examples specific to extxyz
  - Tips for correct format specification

## Scope

- 1 file changed: `skills/dpdata-cli/SKILL.md` (+77 lines)
- All claims traceable to source code
- No behavioral changes, no new dependencies

## Summary by CodeRabbit

* **Documentation**
  * Updated `dpdata-cli` documentation to keep the “Supports 50+ formats” claim while refreshing the showcased supported-format examples, moving away from specific DeePMD variants toward broader references (including `deepmd` and other tools).
  * Added an **Extended XYZ (extxyz) Format Details** section, describing extxyz and its equivalent aliases, clarifying that plain `xyz` is different, and including CLI usage examples and guidance (including `--multi` and type-map usage).


skills/dpdata-cli/SKILL.md
```diff
@@ -173,6 +173,40 @@ Formats may be updated. For the complete and latest list, see:
...
+uvx dpdata data.xyz -i extxyz -O deepmd_data -o deepmd/npy --multi
+
+# Heterogeneous extxyz (mixed compositions) → deepmd/npy/mixed
+uvx dpdata data.xyz -i extxyz -O deepmd_data -o deepmd/npy/mixed --multi
+
+# Convert extxyz to VASP POSCAR (single-frame input only)
+uvx dpdata data.xyz -i extxyz -O POSCAR -o vasp/poscar
+
+# Convert with explicit type map (--multi for multi-frame files)
+uvx dpdata data.xyz -i extxyz -O deepmd_data -o deepmd/npy -t H C N O --multi
```
njzjz: **[P2] Keep the explicit type-map example on a path that applies it**

With `--multi`, `convert()` calls `MultiSystems.from_file(..., type_map=...)`, but `MultiSystems.from_file` creates the container without that map and the extxyz per-frame loader returns each parsed dictionary unchanged. The later `sort_atom_names()` therefore infers an order instead of honoring `-t`. I reproduced the documented form with `-t C B`: both generated `type_map.raw` files still contained `B` followed by `C`. This silently defeats the example’s purpose and can leave training datasets with an unintended type order.

Until the multi-system loader applies the map, make this explicitly single-frame, where `LabeledSystem` does call `apply_type_map`:

```suggestion
# Convert a single-frame extxyz with explicit type map
uvx dpdata data.xyz -i extxyz -O deepmd_data -o deepmd/npy -t H C N O
```

njzjz (changes requested):
Request changes: the frame-preservation correction is sound, but adding `--multi` made the explicit type-map example ineffective because that extxyz path currently ignores `-t`. The inline suggestion keeps the example on the supported single-frame path. The existing unresolved plain-XYZ species wording should also be addressed.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

PR #1037, opened 2026-07-19
fix(dftbplus): parse all atoms instead of truncating to 4

### LINKED ISSUE #991: [Code scan] DFTB+ parser truncates coordinates and forces to four atoms

The DFTB+ parser only reads coordinate rows while `flag in (3, 4, 5, 6)` and force rows while `flag in (8, 9, 10, 11)`. That limits parsing to four atoms regardless of the actual system size.

Affected code:
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/dftbplus/output.py#L41-L60
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/dpdata/formats/dftbplus/output.py#L61-L72

Current tests only cover ammonia, which has four atoms:
https://github.com/deepmodeling/dpdata/blob/a7a50bfc096e1a65470af6f1c48152dc0275ec1c/tests/test_dftbplus.py#L17-L47

Why this is a bug:

- A 5-atom or larger DFTB+ input can only contribute the first four coordinate rows.
- A 5-atom or larger DFTB+ output can only contribute the first four force rows.
- The parser then returns truncated arrays or trips the coordinate/force shape assertion instead of reading all atoms.

The parser should determine the number of geometry/force rows from the file content instead of using fixed flag values for a 4-atom fixture.

abhi-0203:
## Summary

The DFTB+ parser hardcoded flag checks to only read 4 coordinate rows () and 4 force rows (). This silently truncated any system with more than 4 atoms.

Closes #991

## Changes

- Read the atom count from the GenFormat header line (e.g., ) and use it to determine how many coordinate rows to parse
- For forces, parse all consecutive rows after 'Total Forces' until a non-force line is encountered
- Added test with 6-atom system to verify the fix

## Testing

- All existing ammonia (4-atom) tests pass
- New 6-atom test passes
- Linter passes

## Root Cause

The original code used hardcoded flag ranges ( and ) which limited parsing to exactly 4 rows regardless of the actual system size.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DFTB+ output parsing for systems with varying atom counts.
  * Correctly reads all atomic coordinates and force rows, including configurations with more than four atoms.
  * Prevents parsing errors when force sections end with non-numeric lines.

* **Tests**
  * Added coverage for six-atom configurations, including symbols, coordinates, forces, and total energy validation.


njzjz:
Please apply the suggestions proposed by coderabbitai. Thanks!

PR #1044, opened 2026-07-25
docs(formats): expand generated format reference

njzjz:
## Summary

- Add format-owned overviews and NumPy-style parameter documentation for all 51 registered format implementations, 97 conversion methods, and 318 conversion parameters across 83 aliases.
- Rework generated format pages to show the preferred and equivalent aliases, format descriptions, data-model-specific examples, exact public signatures, explicit parameters, and `**kwargs` options; make generation independent of the caller working directory.
- Expand the supported-formats landing page with data-model guidance, Python and CLI conversion examples, format-specific parameter examples, and a description column.
- Add documentation contract tests, enable Ruff D205 while retaining D417 enforcement, and make regular Sphinx validation possible without the optional JupyterLite toolchain.

## Validation

- `ruff format dpdata docs/conf.py docs/make_format.py tests/test_format_docs.py`
- `ruff check dpdata/ docs/conf.py docs/make_format.py tests/test_format_docs.py`
- `python -m unittest discover` — 2238 tests passed, 28 skipped
- `python -m unittest discover -s tests -p 'test_format_docs.py'`
- `dpdata --help`
- `dpdata --version`
- `DPDATA_DOCS_SKIP_JUPYTERLITE=1 sphinx-build -a -b html . _build/html` — succeeded; 1650 existing API/autosummary warning lines remain, with 0 warnings from `docs/formats.rst` or `docs/formats/*.rst`

## Notes

The default JupyterLite build still requires micromamba. `DPDATA_DOCS_SKIP_JUPYTERLITE=1` skips only that optional extension so contributors can validate the regular Sphinx pages. Existing repository-wide API/autosummary warnings remain outside this documentation-focused change.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

## Summary by CodeRabbit

* **Documentation**
  * Expanded supported-format guidance, including how aliases map to implementations, authoritative “supported conversions,” and richer Python/CLI conversion examples.
  * Improved format reference generation with deterministic, clearer conversion documentation and “quick examples.”
  * Refreshed and expanded docstrings across many format adapters (inputs/outputs/parameters), plus wording tweaks for specific readers.

* **Bug Fixes**
  * Made documentation builds more reliable and updated external documentation links; default docs language set to English; JupyterLite can be skipped.

* **Tests**
  * Added automated checks that all registered formats and conversion methods include required parameter documentation.

* **Changes**
  * Clarified the SQM input format as write-only.


dpdata/utils.py
```diff
@@ -175,15 +179,15 @@ def open_file(file: FileType, *args, **kwargs) -> Generator[io.IOBase, None, Non
     ----------
     file : file object or file path
         A file object or a file path.
+    *args : list
+        Positional arguments passed to :func:`open` for path inputs.
+    **kwargs : dict
+        Keyword arguments passed to :func:`open` for path inputs.
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402: `*args` is now documented as `tuple`, and the `ValueError` description reads "If file is neither a file object nor a file path."

dpdata/plugins/abacus.py
```diff
@@ -20,7 +20,30 @@
...
+            Input ``STRU`` file.
+        **kwargs : dict
+            Additional format arguments accepted for API compatibility.
+
+        Returns
+        -------
+        dict
+            System data, including optional ``move`` and magnetic fields when
+            present in the file.
+        """
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402. `get_frame_from_stru` calls `os.path.isfile(stru)` and `open(stru)`, so the type is now `str or os.PathLike`.

dpdata/plugins/amber.py
```diff
@@ -13,6 +13,14 @@
 
 @Format.register("amber/md")
 class AmberMDFormat(Format):
+    """AMBER molecular-dynamics trajectory and label files.
+
+    Coordinates and topology are read from ``.nc`` and ``.parm7`` files.
+    Labeled loading additionally combines optional ``.mdfrc``, ``.mden``,
+    and ``.mdout`` files for forces, energies, and box information. The
+    ``parmed`` optional dependency is required.
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402. `read_amber_traj` opens `mdfrc_file` unconditionally under `if labeled:`, while energies come from `.mden` or `.mdout` (`RuntimeError` if neither exists). The class docstring and `mdfrc_file`/`mden_file`/`mdout_file` entries now say so.

dpdata/plugins/amber.py
```diff
@@ -21,6 +29,27 @@ def from_system(
...
+        ----------
+        file_name : str, optional
+            Common prefix used to infer ``<prefix>.parm7`` and ``<prefix>.nc``.
+        parm7_file : str, optional
+            Explicit AMBER topology file. Overrides the inferred path.
+        nc_file : str, optional
+            Explicit NetCDF trajectory file. Overrides the inferred path.
+        use_element_symbols : list[str] or bool, optional
+            Element-symbol information forwarded to the AMBER reader when
+            atom names cannot be inferred unambiguously from the topology.
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402 at both sites. Now documented as `list[int] or str, optional` — a list of atom indexes, or an AMBER mask string, matching `read_amber_traj`.

dpdata/plugins/qe.py
```diff
@@ -8,8 +8,33 @@
...
+        Parameters
+        ----------
+        file_name : str or os.PathLike
+            Common prefix of the CP input and trajectory files.
+        begin : int, default=0
+            Index of the first frame to load.
+        step : int, default=1
+            Load every ``step``-th frame.
+        **kwargs : dict
+            Additional format arguments accepted for API compatibility.
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402 by narrowing the documented type, which is the accurate option here.

`qe/cp/traj` builds `file_name + ".in"` so a `Path` raises `TypeError`, and `qe/pw/scf` routes through `dpdata.formats.qe.scf.get_frame`, whose `isinstance(fname, str)` check raises `RuntimeError("invalid input")` for a `Path`. Both now document `str`. `qe/pw/scf` also accepts a two-element `[input_file, output_file]` list (exercised in `tests/test_qe_pw_scf.py`), which is now documented too. Making path-like inputs actually work is a behavior change better kept out of this docs-only PR.

dpdata/plugins/xyz.py
```diff
@@ -16,14 +16,30 @@
...
+
+    Plain XYZ stores element symbols and Cartesian coordinates but no cell or
+    labels. dpdata therefore treats it as nonperiodic and assigns a placeholder
+    cell. Use ``extxyz`` when energies, forces, virials, or multiple chemical
+    formulas must be preserved.
 
     Examples
     --------
     >>> s.to("xyz", "a.xyz")
     """
```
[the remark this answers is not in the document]
njzjz: Fixed in e9a0402, with `import dpdata` and an explicit `System` construction before `to()`.

dpdata/plugins/deepmd.py
```diff
@@ -202,6 +296,20 @@ def mix_system(self, *system, type_map, atom_numb_pad=None, **kwargs):
...
+        directory : str or os.PathLike
+            Root directory containing one or more mixed datasets.
+        **kwargs : dict
+            Additional format arguments forwarded when each dataset is read.
+
+        Returns
+        -------
+        list[str]
+            Directories containing ``type_map.raw`` and mixed-type data.
+        """
```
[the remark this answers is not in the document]
njzjz: Not applicable — `DeePMDMixedFormat.from_multi_systems` contains no `yield`. It builds `sys_dir = []`, appends inside `os.walk`, and ends with `return sys_dir`, so `Returns: list[str]` is correct as written.

The generator siblings are `DeePMDHDF5Format.from_multi_systems` and `DeePMDHDF5MixedFormat.from_multi_systems`, and both are already documented with a `Yields` section.

njzjz:
Also addressed the two CodeRabbit comments that could not be posted inline, in e6f267d:

**`dpdata/plugins/deepmd.py` (outside diff range, 492-507)** — confirmed against `MultiSystems.from_fmt_obj` (`dpdata/system.py:1412-1416`), which sends each yielded group to `LabeledSystem().from_fmt_obj` by default and to `System().from_fmt_obj` only when `labeled=False`. `DeePMDHDF5Format.from_multi_systems` now documents both consumers instead of naming only the unlabeled path.

**`docs/make_format.py` (nitpick, 522-533)** — deduplicated, but via a union rather than the suggested per-method lookup: `hidden_doc_parameters = {"*args"}.union(*internal_parameters.values())`. Narrowing to `internal_parameters.get(method, ...)` would leak `mol`/`rdkit_mol` into writer pages backed by a shared `to_system` implementation, which is the caveat your own note raised. The derived set is asserted identical to the previous literal, so generated pages are unchanged.

docs/make_format.py
```diff
@@ -166,73 +402,71 @@ def generate_sub_format_pages(formats: dict):
...
                             method_classes[method],
-                            aa.replace("/", "_").replace(".", ""),
+                            alias.replace("/", "_").replace(".", ""),
                             sig,
                         )
                     )
                     buff.append("""   :noindex:""")
                 buff.append("")
-                if docstring is None or method not in format.__dict__:
+                if docstring is None:
```
wanghan-iapcm, line 522: Dropping the `method not in format.__dict__` guard from #516 means inherited base-class implementations now surface their plugin-author docstrings on user-facing pages.

`check_supported()` advertises `from_multi_systems`/`to_multi_systems` whenever `MultiMode != NotImplemented`, so `getattr` resolves to `dpdata/format.py`'s base methods, whose text reads "Implement MultiSystems.from that converts from this format to MultiSystems. By default, this method follows MultiMode..." -- plus an internal `**kwargs : keyword arguments that will be passed from the method` row.

Regenerating on master vs this branch: master emits ``Convert this format to :class:`MultiSystems`.`` in all these places; this branch leaks the internal text in 5 (`DeePMDCompFormat.rst` 285/333, `DeePMDRawFormat.rst` 266/314, `DeePMDMixedFormat.rst` 226). Same applies to the `to_` branch at line 522.

Could you restore the `__dict__` check, keeping the `to_labeled_system` fallback that `get_method_obj` now handles?

docs/make_format.py
```diff
@@ -130,34 +133,267 @@ def check_supported(fmt: Format):
...
+    for parameter in get_user_parameters(method_obj, method):
+        if parameter.name in location_parameters:
+            arguments.append('"output_path"')
+            has_location = True
+        elif parameter.default is Parameter.empty:
+            arguments.append(f"{parameter.name}={get_parameter_example(parameter)}")
+    return arguments, has_location
+
+
+def append_quick_examples(
```
wanghan-iapcm, line 310: These are advertised as copyable, but for write-only formats the write example names a variable the block never defines, so pasting raises `NameError`.

11 of 51 generated pages are affected: `ListFormat`, `Py3DMolFormat`, `VASPStringFormat`, `PSI4InputFormat`, `SQMINFormat`, `PyMatgenCSEFormat`, plus `DeePMDMixedFormat` and `DeePMDHDF5MixedFormat` (these two define only `systems` from `MultiSystems.from_file`, then call `system.to(...)`), and undefined `atoms`/`molecule`/`structure` on the ASE and pymatgen pages.

This is the same defect you fixed for the hand-written `xyz` example in e9a0402 -- worth fixing in the generator too, either by emitting a construction stub or marking these blocks as non-runnable.

docs/make_format.py
```diff
@@ -130,34 +133,267 @@ def check_supported(fmt: Format):
...
+    "PwmatOutputFormat": "pwmat/output",
+    "QuipGapXYZFormat": "extxyz",
+    "VASPPoscarFormat": "vasp/poscar",
+}
+
+read_source_overrides = {
+    "AmberMDFormat": '"trajectory_prefix"',
+    "ASEStructureFormat": "atoms",
+    "CP2KAIMDOutputFormat": '"calculation_directory"',
+    "DFTBplusFormat": '("geometry.gen", "detailed.out")',
```
wanghan-iapcm: This points users at the wrong file. `read_dftb_plus` scans for `line.startswith("Geometry")` and parses the brace block, i.e. it needs the DFTB+ HSD input, not a standalone GenFormat file -- the repo's own fixture `tests/dftbplus/dftb_pin.hsd` starts with `Geometry = GenFormat {`. An actual `.gen` file (which begins with the atom-count line) raises `ValueError("GenFormat Geometry block not found in DFTB+ input")`.

Suggest `("dftb_in.hsd", "detailed.out")`.

dpdata/plugins/pwmat.py
```diff
@@ -54,8 +84,31 @@ def from_labeled_system(
 @Format.register("pwmat/atom.config")
 @Format.register("pwmat/final.config")
 class PwmatAtomconfigFormat(Format):
+    """PWmat ``atom.config`` or ``final.config`` structure file.
+
+    `PWmat <https://www.pwmat.com/>`_ is a plane-wave DFT software using GPU
+    acceleration.
+
+    The format stores a cell and one atomic configuration. Reading and writing
```
wanghan-iapcm: "Reading and writing normalize the cell to dpdata's lower-triangular convention" is only true for reading. `from_system` carries `@Format.post("rot_lower_triangular")`, but `to_system` has no decorator and `dpdata/formats/pwmat/atomconfig.py::from_system_data` writes `system["cells"][f_idx]` verbatim.

Suggest dropping "and writing".

wanghan-iapcm (commented):
Nice cleanup -- the cwd-independent generation and the per-alias signature fix are real improvements. A few accuracy issues in the generated output, found by regenerating the pages against master and diffing.

docs/make_format.py
njzjz: Fixed in `3d18de1`. Generated conversion docs now use a method docstring only when the format implements that method itself. The intentional `to_labeled_system` fallback to an implementation-owned `to_system` docstring remains. Regeneration confirms that the inherited `MultiMode` plugin-author text is gone, and the full 2,238-test suite passes.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

docs/make_format.py
njzjz: Fixed in `3d18de1`. The generator now defines write-side dpdata objects when no read example created them, and it emits concrete ASE/pymatgen setup objects before those read examples. Regenerated write-only, mixed, ASE, and pymatgen pages no longer reference undefined variables.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

docs/make_format.py
njzjz: Fixed in `3d18de1`. The DFTB+ quick example now uses `("dftb_in.hsd", "detailed.out")`, matching the HSD `Geometry = GenFormat` block consumed by `read_dftb_plus`.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

dpdata/plugins/pwmat.py
njzjz: Fixed in `3d18de1`. The PWmat atom.config overview now states that reading normalizes the cell and no longer claims that writing does so.

Coding agent: Codex
Model: gpt-5.6-sol
Reasoning effort: xhigh

